PR #2, opened 2015-07-25
Generate and insert function py signatures at start of docstring

justinfx:
I thought I would try something simple, just to get into this code base. I use cython a lot, and one feature that is really neat is that it can generate a function signature and insert it into the docstring for you. So I threw together roughly that functionality here.

```go
// GetHelloFor generates a hello msg for a person
func GetHelloFor(person string) string {
	return fmt.Sprintf("Hello, %s!", person)
}
```

```
In [2]: print pytest.GetHelloFor.__doc__
GetHelloFor(str person) str

GetHelloFor generates a hello msg for a person
```

### DIFF
--- a/bind/gencpy.go
+++ b/bind/gencpy.go
@@ -86,7 +86,7 @@ func (g *cpyGen) gen() error {
 
 		case *types.Func:
 			funcs = append(funcs, cpyFunc{typ: obj})
-			docs[obj.Name()] = g.pkg.getDoc(obj)
+			docs[obj.Name()] = g.getPyDoc(obj)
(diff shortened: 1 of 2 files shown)


sbinet:
thanks a lot!
merged in at 06ec70c

justinfx:
My pleasure. I might be inclined to do some more, but since this is such an early state, I would hate to work on the same feature as someone else. Maybe you could throw some feature tickets on here so that I don't look at something in progress?

sbinet:
done. just claim those you like to work on.

Issue #60, opened 2015-09-02
Limit code scanning to package/elements to be exported


kellrott:
I've created an example (https://github.com/kellrott/gopy/tree/wrapper/_examples/wrapper) with two packages, `wrapper` and `wrapper_python`. 
`wrapper` has several features that gopy doesn't support (pointers to structures, channels) and is imported by `wrapper_python`, which simplifies and hides the interaction with those elements. 
When running the command to bind `wrapper_python`
```
gopy bind github.com/go-python/gopy/_examples/wrapper/python
```
I get the error:
```
panic: could not look-up "*github.com/go-python/gopy/_examples/wrapper.RealStruct"!


goroutine 1 [running]:
github.com/go-python/gopy/bind.(*symtab).addMapType(0x823df3ce0, 0x8205d6320, 0x8820774278, 0x8205d65f0, 0x8820774218, 0x8205d0660, 0x88, 0x823e5bf40, 0x4a, 0x823e5bef0, ...)
	/Users/ellrott/workspaces/golang/src/github.com/go-python/gopy/bind/symtab.go:497 +0xa35
github.com/go-python/gopy/bind.(*symtab).addType(0x823df3ce0, 0x8820774278, 0x8205d65f0, 0x8820774218, 0x8205d0660)
	/Users/ellrott/workspaces/golang/src/github.com/go-python/gopy/bind/symtab.go:440 +0x326
github.com/go-python/gopy/bind.(*symtab).addStructType(0x823df3ce0, 0x820647bd0, 0x8820774278, 0x823e5b7c0, 0x88207b8510, 0x8205cc8a0, 0x1108, 0x823e5be00, 0x45, 0x823f50a80, ...)
	/Users/ellrott/workspaces/golang/src/github.com/go-python/gopy/bind/symtab.go:575 +0x27a
github.com/go-python/gopy/bind.(*symtab).addType(0x823df3ce0, 0x8820774278, 0x823e5b7c0, 0x88207b8510, 0x8205cc8a0)
	/Users/ellrott/workspaces/golang/src/github.com/go-python/gopy/bind/symtab.go:383 +0xfd6
github.com/go-python/gopy/bind.(*symtab).addPointerType(0x823df3ce0, 0x820647bd0, 0x8820774278, 0x823e5b7c0, 0x8820774248, 0x823e41fb0, 0x8, 0x823e5bdb0, 0x46, 0x823f50980, ...)
[...]

PR #61, opened 2015-09-03
Ignore private fields in structs

kellrott:
In response to issue #60 
This patch adds a check so that the types of private fields aren't exported to the wrapper. This is useful if writing a 'wrapper' package that buffers the python code from complex GO concepts that aren't yet wrapped for python (like pointers and channels). the code `_examples/wrapper/real_code.go` and `_examples/wrapper/python/wrapper_code.go`  can be compiled with the command `gopy bind github.com/go-python/gopy/_examples/wrapper/python`, and the functions manipulate GO classes gopy doesn't yet support, but they are private, and not exported, so now the code will compile.

### DIFF
--- a/bind/symtab.go
+++ b/bind/symtab.go
@@ -99,6 +99,10 @@ type symbol struct {
 	pychk string
 }
 
+func isPrivate(s string) bool {
+	return (strings.ToLower(s[0:1]) == s[0:1])
+}
+
 func (s symbol) isType() bool {
 	return (s.kind & skType) != 0
 }
@@ -569,6 +573,7 @@ func (sym *symtab) addStructType(pkg *types.Package, obj types.Object, t types.T
 	kind |= skStruct
 	pybuf := make([]string, 0, typ.NumFields())
 	for i := 0; i < typ.NumFields(); i++ {
+		if isPrivate(typ.Field(i).Name()) { continue }
 		ftyp := typ.Field(i).Type()
 		fsym := sym.symtype(ftyp)
 		if fsym == nil {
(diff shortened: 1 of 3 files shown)


bind/symtab.go
```diff
@@ -569,19 +573,21 @@ func (sym *symtab) addStructType(pkg *types.Package, obj types.Object, t types.T
 	kind |= skStruct
 	pybuf := make([]string, 0, typ.NumFields())
 	for i := 0; i < typ.NumFields(); i++ {
-		ftyp := typ.Field(i).Type()
-		fsym := sym.symtype(ftyp)
-		if fsym == nil {
-			sym.addType(typ.Field(i), ftyp)
-			fsym = sym.symtype(ftyp)
+		if ! isPrivate(typ.Field(i).Name()) {
```
sbinet: inverse the condition and exit early (less indentation)

_examples/wrapper/python/wrapper_code.go
```diff
@@ -0,0 +1,24 @@
+// Copyright 2015 The go-python Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package wrapper_python
```
sbinet: as you named the package `wrapper_python`, please rename the directory holding that file to `wrapper_python`.
(or `pywrap` ? `pywrapper` ?)

sbinet:
thanks. LGTM.

one small nitpick though: I strive to follow Go's workflow and coding/commit rules.
could you amend your commit message to be the form of:
```
<name of component being modified>: <short summary>

<(optional) longer explanation of what the change is about. possibly many lines.>

<Fixes/Updates #some-issue>
```

e.g.:

```
bind: ignore private fields in structs

This CL addresses the issue of bla-bla.
So, only handle some foo to bar the qux.

Fixes #60.
```

again, thanks a lot.

bind/symtab.go
kellrott: if isPrivate(typ.Field(i).Name()) { continue }

?

sbinet: yep.

kellrott:
Requested changes should now be in PR

sbinet:
pushed e40a253

PR #62, opened 2015-09-08
Casting struct arguments to method calls.

kellrott:
This is a first attempt to properly cast struct arguments to method calls. The test example would be `func (this S) MethodTest(item S1) {}` from the structs.go example.


sbinet:
LGTM.
Closed with 76adcee

PR #76, opened 2016-01-04
Increment reference count for results in genMethodBody.

lhartung:
I think this was a small omission.  `genFuncBody` generates calls to `cgopy_incref` for all of the return values that require reference counts, but `genMethodBody` was missing that logic.  This causes a panic with the message "cgopy: decref untracked object" for Go methods that return objects when those objects are eventually deleted.

### DIFF
--- a/bind/gengo.go
+++ b/bind/gengo.go
@@ -451,6 +451,13 @@ func (g *goGen) genMethodBody(s Struct, m Func) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


sbinet:
LGTM.

could you please send a PR and add yourself to go-python/license (AUTHORS+CONTRIBUTORS) ?

sbinet:
needs go-python/license#3

Issue #85, opened 2017-05-05
bind: make sure GODEBUG=cgocheck=0 for Go>=1.6


sbinet:
right now, `gopy` isn't following the "new" stricter `cgo` rules.
make sure that (for Go>=1.6) the `GODEBUG` environment variable contains `cgocheck=0` so the generated python module won't crash.

if the environment variable is not set, print out what to do, and continue execution.

corona10:
I am working on this issue 👍

sbinet:
Closed by go-python/gopy#91

PR #89, opened 2017-05-06
bind: make sure GODEBUG=cgocheck=0 for Go>=1.6

corona10:
https://github.com/go-python/gopy/issues/85

1. Check go version.
2. If go version is great equal than 1.6 then check `cgocheck=1` is set, if not raise the error.

### DIFF
--- a/cmd_bind.go
+++ b/cmd_bind.go
@@ -6,14 +6,17 @@ package main
 
 import (
 	"fmt"
+	"github.com/gonuts/commander"
+	"github.com/gonuts/flag"
 	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
-
-	"github.com/gonuts/commander"
-	"github.com/gonuts/flag"
+	"regexp"
+	"runtime"
+	"strconv"
+	"strings"
 )
 
 func gopyMakeCmdBind() *commander.Command {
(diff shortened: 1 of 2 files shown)


corona10:
@sbinet 
This is the Pull Request for GSoC 2017.
Coding style might be different as you expected.
Please point it out. This is why community bonding period exists!.
Thanks!

cmd_bind.go
```diff
@@ -6,14 +6,17 @@ package main
 
 import (
 	"fmt"
+	"github.com/gonuts/commander"
```
sbinet: please keep the "stdlib" / "3rd-party" separation of imports.

https://godoc.org/golang.org/x/tools/cmd/goimports should do that for you.

cmd_bind_test.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2015 The go-python Authors.  All rights reserved.
```
sbinet: for new files, change the copyright to '2017'

sbinet:
thanks for doing this.

it's actually not exactly what I had in mind (but I've expressed myself poorly)
the check must happen at runtime, when the generated python module is imported by the python VM.

so when `gopy` is compiled with a Go toolchain >= 1.6, it should:
- inject in the generated cgo file an `init()` function
- the `init()` function would then check if the `GODEBUG` environment variable has the correct value
- print a warning (to `os.Stderr`) if that's not the case

corona10:
@sbinet 
Thank you for reviewing my PR.
I understand what you want to do.
I will write code about that solution.
Then I will close this PR and open the new PR for this issue.

corona10:
@sbinet 
One more question about this issue.
Is there any good tip for detecting Go version is >= 1.6? on python runtime?

corona10:
ah okay, I can do it. I already wrote the solutions on the code XD.

sbinet:
Yep. It's basically putting the Go code you write inside an  'init' function inside the generated code :)

sbinet:
(no need to close this PR, just push your new code inplace)

PR #91, opened 2017-05-06
bind: make sure GODEBUG=cgocheck=0 for Go>=1.6

corona10:
#85 
When gopy is compiled with a Go toolchain >= 1.6
It inject in the generated cgo file an init() function
the init() function would then check if the GODEBUG environment variable has the correct value
print a warning (to os.Stderr) if that's not the case

Here is the mainly injected C codes.

```` c
        /* make sure GODEBUG=cgocheck=0 for Go>=1.6 */
        cgo_check();

        /* make sure Cgo is loaded and initialized */
        cgo_pkg_hi_init();
````

### CODE REVIEW

**Comment by corona10** [comment]
@sbinet 
As I mentioned before 
Coding style might be different as you expected. (This time will be C code style. :))
Please point it out.
Thanks!

**Comment by corona10** [comment]
@sbinet If you want the implementation with [CAPI](https://docs.python.org/2/c-api/string.html)
Then I will update it!

**Comment by sbinet** [comment]
IMHO it would be better to have the check done in the generated .go file instead of the one using the Python capi because then we won't have to rewrite it when we'll migrate the (default) Python backend to cffi-based generated code.

**Comment by corona10** [comment]
@sbinet You are right that would be the better solution. It looks like I should work on /bind/gengo.go to generate it. right?

**Comment by sbinet** [comment]
yep.

**Comment by corona10** [comment]
@sbinet 
Updated it. I am now more familiar with the codebases which you wrote.
I wish that I would not interrupt your comfortable weekend. :)

**Comment by corona10** [comment]
@sbinet 
I could merge `cgo_check()` and `getCgoCheck` into `cgo_check()`. If you would like to.

**Comment by corona10** [comment]
Ready to review!

File: bind/gencpy.go [hunk 1]
```diff
@@ -259,6 +260,13 @@ func (g *cpyGen) gen() error {
 	g.impl.Indent()
 	g.impl.Printf("PyObject *module = NULL;\n\n")
 
+	version := runtime.Version()
```
sbinet: instead of having this check being made during the `gencpy` step, put it during the `gengo` step.

bind/gengo.go
```diff
@@ -111,6 +114,28 @@ func cgopy_decref(ptr unsafe.Pointer) {
 	refs.Unlock()
 }
 
+//export cgo_check
```
sbinet: and instead of creating a `cgo_check` function that is then cgo-exported to be called from the `gencpy` step, generate an `init()` function that will *only* be inserted inside the generated code when the Go toolchain generating that file is `>=1.6`.

this minimizes the number of call sites to maintain for consistency. (and slightly simplifies the call chain.)

bind/gengo.go
```diff
@@ -111,6 +114,28 @@ func cgopy_decref(ptr unsafe.Pointer) {
...
+func cgo_check() {
+	godebug := os.Getenv("GODEBUG")
+	cgocheck := -1
+	var err error
+	if godebug != "" {
+                for _, option := range strings.Split(godebug, ",") {
+                        if strings.HasPrefix(option, "cgocheck=") {
```
sbinet: instead of checking if it has the correct prefix, check whether it isn't there and `continue`.
this decreases the amount of indentation.

bind/gengo.go
```diff
@@ -111,6 +114,28 @@ func cgopy_decref(ptr unsafe.Pointer) {
...
+                                cgocheck, err = strconv.Atoi(option[9:])
```
sbinet: instead of `option[9:]`, replace with:
`option[len(prefix):]`

where `prefix` is defined as:

```go
const prefix = "cgocheck="
```

bind/gengo.go
```diff
@@ -111,6 +114,28 @@ func cgopy_decref(ptr unsafe.Pointer) {
...
+                                if err != nil {
+					cgocheck = -1
+					fmt.Fprintf(os.Stderr, "Invalid cgocheck: %%s\n", option)
```
sbinet: the warning should be prefixed with `"gopy:"`

e.g.:

```go
fmt.Fprintf(os.Stderr, "gopy: invalid cgocheck value %%q (expected an integer)\n", option)
```

bind/gengo.go
```diff
@@ -111,6 +114,28 @@ func cgopy_decref(ptr unsafe.Pointer) {
...
+                                }
+                        }
+                }
+        }
+
+	if cgocheck != 0 {
+		fmt.Fprintf(os.Stderr, "GODEBUG=cgocheck=0 should be set for Go>=1.6\n")
```
sbinet: ditto: please prefix the warning message with `gopy: `

bind/utils.go
```diff
@@ -148,3 +150,15 @@ func getPkgConfig(vers int) (string, error) {
 
 	return pkgcfg, nil
 }
+
+func getGoVersion(version string) (int64, int64, error) {
+	version_regex := regexp.MustCompile(`^go((\d+)(\.(\d+))*)`)
+	match := version_regex.FindStringSubmatch(version)
+	if match == nil {
+		return -1, -1, fmt.Errorf("Invalid Go version information: %s", version)
```
sbinet: `s/Invalid/invalid`

also, perhaps replace the `%s` verb with `%q`

bind/utils_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+package bind
+
+import (
+	"errors"
+	"testing"
+)
+
+func TestGetGoVersion(t *testing.T) {
+	var versionTests = []struct {
```
sbinet: no need to create a named variable for this list of tests.
just create that list inline of the `for` loop.

people are used to table-driven tests.

_i.e._:

```go
for _, tt := range []struct {
    info string
    ...
}{
   {"go1.5", 1, 5, nil},
   ...
}{
   major, minor, err := getGoVersion(tt.info)
   ...
}
```

corona10:
@sbinet Huh A lot of things!. 
To summarize the key, is it that you want to call cgocheck() in init() only when the Go version is 1.6 or higher? Of course, should this code be generated only when Go version is 1.6 or higher?

sbinet:
yes. `gengo` should only generate the version check when `version>=1.6`.
this way, the generated code doesn't have to import `runtime`.

one way would be to modify the `gengo` preamble to have at:
https://github.com/corona10/gopy/blob/c8e722b67fb59fe62e7fd6e79a1de3b1c046c320/bind/gengo.go#L149

another `%s` field that would be just `""` when `version <1.5` and the call to `cgopy_check()` otherwise.
(and modify `func (*goGen) genPreamble()` accordingly).

bind/utils_test.go
corona10: @sbinet 
Thank you for letting me know the better way.
I learn a lot from you.

corona10:
@sbinet 
Thanks, @sbinet  I got it. I will use cleaner way.

sbinet:
no worries: the first PR is where we get a lot of the common vocabulary and expectations :)
thanks for sticking with it.

corona10:
@sbinet 
One more question, 
>> this way, the generated code doesn't have to import runtime.

`version := runtime.Version()` should be called anyway to check go version?

sbinet:
only during the generation of the Cgo file.

the Cgo file itself doesn't have to import `runtime` and check the version: the current modus operandi of the `buildmode=shared` is to crash and burn if the Go versions (used at build and used at runtime) differ.
so we can assume and rely on the fact that, at runtime, if the `init()` function has the cgo-check part, then Go version >= 1.6.

we can hope that this assumption won't have to hold very long and have `gopy` support Go version >= 1.6 (sooner than later :P)

corona10:
@sbinet Ready for the review! PTAL

bind/gengo.go
```diff
@@ -8,10 +8,40 @@ import (
...
 )
 
 const (
+	cgo_check_import = `"strconv"
+	"strings"
+	"os"
+`
+	cgo_check= "cgo_check()"
+	cgo_checkDef = `
+func cgo_check() {
```
sbinet: for consistency with other `gopy` internal functions, please call it `cgopy_check`.

corona10: `_copy_check` is looks great? Similar function name convention to  `_cgopy_GoString`.

sbinet: yes, right.
actually, let's be a bit more specific: `_cgopy_CheckGoVersion` :)

corona10:
@sbinet Done!

sbinet:
ok. could you send a PR against [go-python/license](https://github.com/go-python/license), adding yourself to the `AUTHORS` and `CONTRIBUTORS` files?
then, I'll merge the 2 PRs.

thanks a lot.

corona10:
https://github.com/go-python/license/pull/5

Send it.

Issue #48, opened 2015-08-26
gopy: consider support of a CFFI backend


sbinet:
prompted by: https://www.reddit.com/r/Python/comments/3ig3ww/building_python_modules_with_go_15_xpost_from/cugc9ri

it could be interesting to have a `CFFI` backend for `pypy`'s benefit.
maybe simpler to just tap into `cppyy` ?

see https://github.com/jbuberel/buildmodeshared/tree/master/gofrompython for a `CFFI`-based example

corona10:
I am working on this issue.

Issue #87, opened 2017-05-05
gopy,bind: add boilerplate code for cffi support


sbinet:
this means:

- `gopy/gen.go`: add a new `"cffi"` language switch
- `gopy/bind/...`: add a new `cffiGen` type, modeled after `cpyGen`.

corona10:
## What I did.

I summary here for few things that I research about generating cffi support.

For this research, I used `_example/simple.go` for proto-typing.
The first step, Generating go file by `gengo.go`. This is already implemented.
Then gopy should generate `_simple.so`  by `go build -buildmode=c-shared -o _simple.so simple.go`

Then cffigen.go generate this `build.py`.

```` python
from cffi import FFI

ffibuilder = FFI()

ffibuilder.set_source(
    '_gopysimple',
    None,
    extra_objects=["_simple.so"],
)

ffibuilder.cdef("""
typedef signed char GoInt8;
typedef unsigned char GoUint8;
typedef short GoInt16;
typedef unsigned short GoUint16;
typedef int GoInt32;
typedef unsigned int GoUint32;
typedef long long GoInt64;
typedef unsigned long long GoUint64;
typedef GoInt64 GoInt;
typedef GoUint64 GoUint;
typedef float GoFloat32;
typedef double GoFloat64;
typedef struct { const char *p; GoInt n; } GoString;

extern GoString _cgopy_GoString(char* p0);
extern char* _cgopy_CString(GoString p0);
extern void cgo_pkg_simple_init();
extern void cgo_check();
extern GoInt cgo_func_simple_Add(GoInt p0, GoInt p1);
extern void cgo_func_simple_Func();
""")

if __name__ == "__main__":
    ffibuilder.compile()
````

Second, It will also generate `simple.py` in this way.

```` python
from __future__ import absolute_import

import os
from _gopysimple import ffi

here = os.path.dirname(os.path.abspath(__file__))
lib = ffi.dlopen(os.path.join(here, "_simple.so"))

lib.cgo_check()

lib.cgo_pkg_simple_init()

def cffi_func_simple_Func():
    lib.cgo_func_simple_Func()

def cffi_func_simple_Add(p0, p1):
    return lib.cgo_func_simple_Add(p0, p1)

def Func():
    return cffi_func_simple_Func()

def Sum(p0, p1):
    return cffi_func_simple_Add(p0, p1)

````

Finally, gopy will run `python build.py` and then copy files which will be needed for users.

## Problems
- 
   ````c
[...]

PR #93, opened 2017-05-29
gopy,bind: add boilerplate code for cffi support

corona10:
* gopy/gen.go: Add a new "cffi" language switch
* gopy/cmd_bind.go: Reduce to using cmd copy command.
* gopy/bind/gencffi.go: Add a new cffiGen type, modeled after cpyGen.
* gopy/bind/gencffi_cdef.go: Expose CAPI generated by cgo for CFFI.
* gopy/bind/gencffi_func.go: Generate CFFI wrapped functions.
* gopy/gen.go: Generate a CFFI builder python file and a wrapping CFFI python file.
* goby/main_test.go: Add test case for CFFI with simple.go.

Fixes go-python/gopy#48, go-python/gopy#87

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,13 @@ go:
   - 1.6.x
   - 1.7.x
   - 1.8.x
+  - master
 
-sudo: false
+sudo: required
 
 before_install:
+   - sudo apt-get install libffi-dev
+   - sudo pip install cffi
    - export PATH=$HOME/gopath/bin:$PATH
    # temporary workaround for go-python/gopy#83
    - export GODEBUG=cgocheck=0
--- a/bind/bind.go
+++ b/bind/bind.go
@@ -53,6 +53,42 @@ func GenCPython(w io.Writer, fset *token.FileSet, pkg *Package, lang int) error 
 	return err
 }
 
+// GenCFFI generates a CFFI package from a Go package
+// Use 4spaces indentation for Python codes, aka PEP8.
+// b is an io.Writer for a builder python script.
+// w is an io.Writer for a wrapper python script which will be executed by a user.
+//
+// GenCFFI generates a builder python script and a wrapper python script by 3 steps.
+// First, GenCFFI analyze which interfaces should be exposed from the Go package.
+// Second, Then GenCFFI writes a CFFI builder package with writing exposed interfaces
+// Third, The GenCFFI writes a wrapper python script.
+func GenCFFI(b io.Writer, w io.Writer, fset *token.FileSet, pkg *Package, lang int) error {
+	gen := &cffiGen{
+		builder: &printer{buf: new(bytes.Buffer), indentEach: []byte("    ")},
+		wrapper: &printer{buf: new(bytes.Buffer), indentEach: []byte("    ")},
+		fset:    fset,
+		pkg:     pkg,
+		lang:    lang,
+	}
+
+	err := gen.gen()
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(b, gen.builder)
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(w, gen.wrapper)
+	if err != nil {
+		return err
+	}
+
+	return err
+}
+
 // GenGo generates a cgo package from a Go package
 func GenGo(w io.Writer, fset *token.FileSet, pkg *Package, lang int) error {
 	buf := new(bytes.Buffer)
--- a/bind/gencffi.go
+++ b/bind/gencffi.go
@@ -1,0 +1,160 @@
+// Copyright 2017 The go-python Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package bind
+
+import (
+	"go/token"
+)
+
+const (
+	/*
+	   FIXME(corona10): ffibuilder.cdef should be written this way.
+	   ffibuilder.cdef("""
+	   	//header exported from 'go tool cgo'
+	   	#include "%[3]s.h"
+	   	""")
+
+	   * discuss: https://github.com/go-python/gopy/pull/93#discussion_r119652220
+	*/
+	builderPreamble = `import os
+
+from sys import argv
+from cffi import FFI
+
+ffibuilder = FFI()
+ffibuilder.set_source(
+    '%[1]s',
+    None,
+    extra_objects=["_%[1]s.so"],
+)
+
+ffibuilder.cdef("""
+typedef signed char GoInt8;
+typedef unsigned char GoUint8;
+typedef short GoInt16;
+typedef unsigned short GoUint16;
+typedef int GoInt32;
+typedef unsigned int GoUint32;
+typedef long long GoInt64;
+typedef size_t GoUintptr;
+typedef unsigned long long GoUint64;
+typedef GoInt64 GoInt;
+typedef GoUint64 GoUint;
+typedef float GoFloat32;
+typedef double GoFloat64;
+typedef struct { const char *p; GoInt n; } GoString;
+typedef void *GoMap;
+typedef void *GoChan;
+typedef struct { void *t; void *v; } GoInterface;
+typedef struct { void *data; GoInt len; GoInt cap; } GoSlice;
+
+extern GoString _cgopy_GoString(char* p0);
+extern char* _cgopy_CString(GoString p0);
+extern GoUint8 _cgopy_ErrorIsNil(GoInterface p0);
+extern char* _cgopy_ErrorString(GoInterface p0);
+extern void cgopy_incref(void* p0);
+extern void cgopy_decref(void* p0);
+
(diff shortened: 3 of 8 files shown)


.travis.yml
```diff
@@ -8,6 +8,8 @@ sudo: false
 before_install:
    - export PATH=$HOME/gopath/bin:$PATH
 
+install: "pip install --user -r requirements.txt"
```
sbinet: perhaps, instead of using `pip`, one could use `yum` to install `setuptools` and `cffi` ?

sbinet: also, this line `"pip install ..."` should be moved in the `before_install` block.

corona10: @sbinet Ah okay. Actually, I am not familiar with yum install, So I need to find how to install cffi with yum. :)

sbinet: actually, the travis containers are running ubuntu, so that would be `apt-get` :)
ie: `apt-get install python-cffi`

sbinet: or, `sudo apt-get install -y python-cffi` :)

.travis.yml
```diff
@@ -3,9 +3,11 @@ language: go
 go:
   - 1.5.2
 
-sudo: false
+sudo: required
 
 before_install:
+   - sudo apt-get update -y
+   - sudo apt-get install -y python-cffi
```
sbinet: arf. when `sudo: required` is enabled, we get a quite old ubuntu release (`precise` to be precise).
so, well, one will need to use `pip` instead.

corona10:
@sbinet Hmm it's strange. `main_test.go:148: [_examples/simple]: error running python module: signal: segmentation fault (core dumped)` is not happend on local.

The similar issue was written in [stack-overflow](https://stackoverflow.com/questions/34544445/travis-segmentation-fault-but-works-fine-locally).

corona10:
@sbinet It looks like works well on go>=1.6 
https://travis-ci.org/go-python/gopy/jobs/237196400

corona10:
@sbinet Now it works! PTAL

corona10:
@sbinet Now we can test every go versions. (1.5.x - master)

bind/gencffi.go
```diff
@@ -0,0 +1,152 @@
...
+
+from sys import argv
+from cffi import FFI
+
+os.chdir(os.path.dirname(os.path.abspath(__file__)))
+ffibuilder = FFI()
+ffibuilder.set_source(
+    'gopy.%[1]s',
+    None,
+    extra_objects=["gopy/_%[1]s.so"],
```
corona10: The first problem if we do not create an extra directory.
If we generate with `simple.go`, CFFI will generate `simple.py` also.(This flow is uncontrollable.) 
This file name is duplicated. Because we also need to generate `simple.py` also for user importing files.
(I think that user wants to import our generated files as  not `import gosimple` but 'import simple`.)
So that's why I create the extra directory as `gopy` and then put it into. And then we could write wrapper python file the same name as `simple.py'.

The second problem is that It will be little dirty because there are 4 output files. So if we put most of them into `gopy` directory. the user can easily remove them.

bind/gencffi.go
```diff
@@ -0,0 +1,152 @@
...
+const (
+	builderPreamble = `import os
+
+from sys import argv
+from cffi import FFI
+
+os.chdir(os.path.dirname(os.path.abspath(__file__)))
+ffibuilder = FFI()
+ffibuilder.set_source(
+    'gopy.%[1]s',
```
corona10: @sbinet 
This is just related to CFFI library.
The user will import as `import XXXX` not `import gopy.XXXX`

cmd_bind.go
```diff
@@ -126,32 +127,111 @@ func gopyRunCmdBind(cmdr *commander.Command, args []string) error {
...
+
+	err = copyFile(filepath.Join(work, "build_"+pkgname)+".py", filepath.Join(odir, "build_"+pkgname)+".py")
+	if err != nil {
+		return err
+	}
+
+	cmd := exec.Command(
+		"python",
+		filepath.Join(odir, "build_"+pkgname)+".py",
+		"",
```
corona10: Cool! That's awesome!

corona10:
@sbinet I am going to fix not to use `gopy` directory as you commented.

corona10:
@sbinet Please take a look. I changed a way for generating CFFI backend packages.

### Before b6ca49c,
It generated CFFI packages for `XXXX.go` in this way.
``````
output_dir/gopy/_XXXX.so
output_dir/gopy/XXXX.py (CFFI related python script. automatically generated by CFFI).
output_dir/XXXX.py (Generated by gopy genCFFI wrapped python file. User will import this python file.)
``````
* Pros: if we need to use a library or grammar for `output_dir/XXXX.py ` which should be position on the top of python file. It could be easily solved. 
* Cons: Too many output files.

### After b6ca49c,
``````
output_dir/_XXXX.so
output_dir/XXXX.py (CFFI related python script which automatically generated by CFFI + Generated by gopy genCFFI wrapped python file. User will import this python file.)
``````

* Pros: Not too many output directory and files. 
* Cons: if we need to use a library or grammar for `output_dir/XXXX.py ` which should be position on the top of python file. We need to find some kinds of merging policy. Because of Python script which automatically generated by CFFI. (e.g from __future__ import xxx)

I will write a comment for `GenCFFI` function after we choose which way to use.

#### Easy way to see the difference is testing with `simple.go` case for CFFI.

**Comment by corona10** [comment]
@sbinet 
Ready to review.

**Comment by sbinet** [comment]
Ok, thanks.
I'll tackle it this afternoon.

File: bind/bind.go [hunk 6]
```diff
@@ -53,6 +53,44 @@ func GenCPython(w io.Writer, fset *token.FileSet, pkg *Package, lang int) error
 	return err
 }
 
+/**
```
sbinet: the canonical way to add documentation for a Go type, func, var, const, ... is to use `// `.
yeah, even for multi-lines comments.

bind/gencffi.go
```diff
@@ -0,0 +1,149 @@
...
+
+if __name__ == "__main__":
+    ffibuilder.compile()
+    # Remove itself after compile.
+    os.remove(argv[0])
+`
+
+	cffiPreamble = `
+__doc__="""%[1]s"""
+import os
```
sbinet: please add an empty line between `doc` and the first import.

bind/gencffi.go
```diff
@@ -0,0 +1,149 @@
...
+
+here = os.path.dirname(os.path.abspath(__file__))
+lib = ffi.dlopen(os.path.join(here, "_%[2]s.so"))
+
+# python <--> cffi helpers.
+def cffi_cnv_py2c_string(o):
```
sbinet: it might be better to not export all these `cffi_cnv_xxx` function.
or at least, clearly say they shouldn't be used by prepending a `_` in front of the `cffi_cnv_xxx`.

ie:
`def _cffi_cnv_py2c_string(o):`

corona10:
@sbinet 
Thanks for the great code review during the your busy time.
I fixed them by your comment. 
I think that we walked a great step. Again, Thanks man!

bind/gencffi.go
corona10: Done dfd4047

corona10:
@sbinet 
Ready to review.

**Before 01748c6 and dfd4047**
```
>>> dir(simple)
['Add', 'Func', '__builtins__', '__doc__', '__file__', '__name__', '__package__', '_cffi_backend',
 'cffi_cnv_c2py_float32', 'cffi_cnv_c2py_float64', 'cffi_cnv_c2py_int', 'cffi_cnv_c2py_string', 
'cffi_cnv_py2c_float32', 'cffi_cnv_py2c_float64', 'cffi_cnv_py2c_int', 'cffi_cnv_py2c_string',
 'ffi', 'here', 'lib', 'os']
```

**After**
```
>>> dir(simple)
['Add', 'Func', '__builtins__', '__doc__', '__file__', '__name__', '__package__',
 '_cffi_backend', '_helper', 'ffi', 'os']
```

corona10:
@sbinet
 I think that we can remove builder script process by wrapping simple.py by [this way.](https://gist.github.com/corona10/5571727becfecd08eca9768be5b3a48e)
If you are okay. I will upload for new PR after this PR is merged.
WDYT?

sbinet:
ok, could you squash everything into one single commit:

```gopy,bind: add initial boilerplate code for cffi support

This CL adds the initial infrastructure to support CFFI as a backend for gopy.
It also adds the necessary dependencies for Travis-CI to test CFFI.

Updates go-python/gopy#48
Fixes go-python/gopy#87
```

then I'll finally merge this :)

corona10:
Done!

sbinet:
thanks, but I've seen it too late and pushed the `gofmt-check` branch.
could you rebase your branch on top of the new `master` (`7c437e9`).

sorry about that.

corona10:
That's okay. Now it pass `gofmt` :)

sbinet:
thanks for your patience.

PR #95, opened 2017-05-29
gopy: add python-cffi dependency to travis config

sbinet:


corona10:
@sbinet This might be a better way. :)

sbinet:
ok. according to travis-ci documentation:
https://docs.travis-ci.com/user/languages/python/#Travis-CI-Uses-Isolated-virtualenvs

```
Travis CI Uses Isolated virtualenvs #

CI Environment uses separate virtualenv instances for each Python version. System Python is not used and should not be relied on. If you need to install Python packages, do it via pip and not apt.

If you decide to use apt anyway, note that Python system packages only include Python 2.7 libraries on Ubuntu 12.04 LTS. This means that the packages installed from the repositories are not available in other virtualenvs even if you use the –system-site-packages option.
```

so, using `pip` is the recommended way.

live'n'learn...

corona10:
@sbinet We learned something useful information. :)

PR #98, opened 2017-06-07
bind, gencffi: Remove generating builders and support Vars and Consts.

corona10:
- Remove the generating builder python script.
- Now, genCFFI interface is same as the genCpy.
- Support builtin-types Vars and Consts.
- Pass `empty.go`
- Pass `cgo.go`
- Pass `vars.go` + Add set vars test.
- Pass `consts.go`

Updates: go-python/gopy#102

### DIFF
--- a/_examples/vars/test.py
+++ b/_examples/vars/test.py
@@ -7,6 +7,11 @@ from __future__ import print_function
 
 import vars
 
+print("doc(vars):\n%s" % repr(vars.__doc__))
+print("doc(vars.GetV1()):\n%s" % repr(vars.GetV1.__doc__))
+print("doc(vars.SetV1()):\n%s" % repr(vars.SetV1.__doc__))
+
+print("Initial values")
 print("v1 = %s" % vars.GetV1())
 print("v2 = %s" % vars.GetV2())
 print("v3 = %s" % vars.GetV3())
@@ -21,3 +26,24 @@ print("k2 = %s" % vars.GetKind2())
 #print("k3 = %s" % vars.GetKind3())
 #print("k4 = %s" % vars.GetKind4())
 
+vars.SetV1("test1")
+vars.SetV2(90)
+vars.SetV3(1111.1111)
+vars.SetV4("test2")
+vars.SetV5(50)
+vars.SetV6(50)
+vars.SetV7(1111.1111)
+
+vars.SetKind1(123)
+vars.SetKind2(456)
+print("New values")
+print("v1 = %s" % vars.GetV1())
+print("v2 = %s" % vars.GetV2())
+print("v3 = %s" % vars.GetV3())
+print("v4 = %s" % vars.GetV4())
+print("v5 = %s" % vars.GetV5())
+print("v6 = %s" % vars.GetV6())
+print("v7 = %s" % vars.GetV7())
+
+print("k1 = %s" % vars.GetKind1())
+print("k2 = %s" % vars.GetKind2())
--- a/bind/bind.go
+++ b/bind/bind.go
@@ -55,16 +55,13 @@ func GenCPython(w io.Writer, fset *token.FileSet, pkg *Package, lang int) error 
 
 // GenCFFI generates a CFFI package from a Go package
 // Use 4spaces indentation for Python codes, aka PEP8.
-// b is an io.Writer for a builder python script.
 // w is an io.Writer for a wrapper python script which will be executed by a user.
 //
-// GenCFFI generates a builder python script and a wrapper python script by 3 steps.
-// First, GenCFFI analyze which interfaces should be exposed from the Go package.
-// Second, Then GenCFFI writes a CFFI builder package with writing exposed interfaces
-// Third, The GenCFFI writes a wrapper python script.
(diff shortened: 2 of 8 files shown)


corona10:
Ready for review.

sbinet:
please rebase on top of the new `master` (`6456a22`).
thanks.

corona10:
Yeah done!

corona10:
@sbinet Ready to review :) PTAL

_examples/vars/test.py
```diff
@@ -7,6 +7,7 @@
 
 import vars
 
+print("Before set values")
```
sbinet: `s/Before set values/Initial values/`

_examples/vars/test.py
```diff
@@ -21,3 +22,24 @@
...
+vars.SetV2(90)
+vars.SetV3(1111.1111)
+vars.SetV4("test2")
+vars.SetV5(50)
+vars.SetV6(50)
+vars.SetV7(1111.1111)
+
+vars.SetKind1(123)
+vars.SetKind2(456)
+print("After set values")
```
sbinet: `s/After set values/New values/`

bind/bind.go
```diff
@@ -55,16 +55,13 @@ func GenCPython(w io.Writer, fset *token.FileSet, pkg *Package, lang int) error
...
-// b is an io.Writer for a builder python script.
 // w is an io.Writer for a wrapper python script which will be executed by a user.
 //
-// GenCFFI generates a builder python script and a wrapper python script by 3 steps.
+// GenCFFI generates a wrapper python script by 2 steps.
 // First, GenCFFI analyze which interfaces should be exposed from the Go package.
-// Second, Then GenCFFI writes a CFFI builder package with writing exposed interfaces
-// Third, The GenCFFI writes a wrapper python script.
-func GenCFFI(b io.Writer, w io.Writer, fset *token.FileSet, pkg *Package, lang int) error {
+// Second, The GenCFFI writes a wrapper python script.
```
sbinet: `s/Second, The GenCFFI/Then, GenCFFI/`

sbinet: also:
`s/First, GenCFFI analyze/First, GenCFFI analyzes/`

bind/gencffi_func.go
```diff
@@ -32,6 +32,52 @@ def %[2]s(%[3]s):
...
+	args := sig.Params()
+
+	var funcArgs []string
+	for _, arg := range args {
+		funcArgs = append(funcArgs, arg.Name())
+	}
+	g.wrapper.Printf(`
+# pythonization of: %[1]s.%[2]s 
+def Get%[2]s(%[3]s):
+`,
```
sbinet: perhaps inject the documentation (if any) for the variable we get here?
so when we do `help(mymodule.GetFoo)` we get the documentation that was attached to the `Foo` variable on the Go side.

that would look like:

```go
g.wrapper.Printf(`
# pythonization of: %[1]s.%[2]s
def Get%[2]s(%[3]s):
    """%[4]s""""
`,
  g.pkg.pkg.Name(),
  o.GoName(),
  strings.Join(funcArgs, ", "),
  o.Doc(),
)
```

bind/gencffi_func.go
```diff
@@ -32,6 +32,52 @@ def %[2]s(%[3]s):
...
+	args := sig.Params()
+
+	var funcArgs []string
+	for _, arg := range args {
+		funcArgs = append(funcArgs, arg.Name())
+	}
+	g.wrapper.Printf(`
+# pythonization of: %[1]s.%[2]s 
+def Set%[2]s(%[3]s):
+`,
```
sbinet: same here.

bind/gencffi_func.go
corona10: @sbinet That's cool. I will do that :)

corona10:
@sbinet Done!

main_test.go
```diff
@@ -540,6 +563,41 @@ v6 = 42
...
+v7 = 1111.1111
+k1 = 123
+k2 = 456
+`),
+	})
+
+	testPkgWithCFFI(t, pkg{
+		path: "_examples/vars",
+		want: []byte(`Initial values
+v1 = v1
```
sbinet: perhaps also add here a test of the docstring?

see:
https://github.com/go-python/gopy/blob/master/_examples/empty/test.py#L10
https://github.com/go-python/gopy/blob/master/main_test.go#L399

corona10: @sbinet Just for the `vars.__doc__`? or for all the Get/Set functions?
Only vars__doc__ is okay but for each of Get/Set functions might be a problem.
The problem of Get/Set functions docstring is that output is different between Cpy2 and CFFI.

````
# CFFI output
doc(vars.GetV1()):
'returns vars.V1'

# CPy2 output
doc(vars.GetV1()):
''
````

sbinet: and for one `Get/Set` pair.

it's OK to have a different reference for CFFI and for CPython-2.
(if there is no issue number for the CPython-2 backend to generate docstrings for `Get/Set`s of vars, I'll add one.)

corona10:
Updated!

corona10:
@sbinet Thank you for reviewing this. :) PTAL

sbinet:
thanks

PR #99, opened 2017-06-08
Gofmt check

sbinet:
This CL adds a test to make sure we always correctly `gofmt` our code (actually, if `goimports` is available, we run it instead of `gofmt`)

### DIFF
--- a/_examples/wrapper/pywrapper/wrapper_code.go
+++ b/_examples/wrapper/pywrapper/wrapper_code.go
@@ -6,6 +6,7 @@ package pywrapper
 
 import (
 	"fmt"
+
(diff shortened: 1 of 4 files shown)


sbinet:
yeah `goimports` is just great (and so convenient!!)
the only thing that `goimports` misses _wrt_ `gofmt` is the `-s` flag:

```sh
$> gofmt -h
[...]
  -s	simplify code
[...]
```

which can, _e.g._, rewrite:
```go
for i, _ := range []int{1,2,3} { ... }
for _ = range someChan { ... }
```

into:
```go
for i := range []int{1,2,3} { ... }
for range someChan { ... }
```

(among other things)

Issue #104, opened 2017-06-15
gencffi: Detect functions returning a Go error


corona10:
* Detect functions returning a Go error and make them pythonic (raising an Exception).
* Add tests for it.

PR #105, opened 2017-06-15
gencffi: Detect functions returning a Go error

corona10:
* Detect functions returning a Go error and make them pythonic (raising an Exception).
* Add tests for it.

Now it only supports 2 return values.
CPy2 and CFFI should support more return values.

Fixes: go-python/gopy#104

### DIFF
--- a/bind/gencffi_func.go
+++ b/bind/gencffi_func.go
@@ -109,6 +109,38 @@ func (g *cffiGen) genFuncBody(f Func) {
 
 	g.wrapper.Printf("_cffi_helper.lib.cgo_func_%[1]s(%[2]s)\n", f.id, strings.Join(funcArgs, ", "))
 
+	if f.err {
+		switch nres {
+		case 1:
+			g.wrapper.Printf("if not _cffi_helper.lib._cgopy_ErrorIsNil(cret):\n")
+			g.wrapper.Indent()
+			g.wrapper.Printf("c_err_str = _cffi_helper.lib._cgopy_ErrorString(cret)\n")
+			g.wrapper.Printf("py_err_str = ffi.string(c_err_str)\n")
+			g.wrapper.Printf("_cffi_helper.lib._cgopy_FreeCString(c_err_str)\n")
+			g.wrapper.Printf("raise RuntimeError(py_err_str)\n")
+			g.wrapper.Outdent()
+			g.wrapper.Printf("return\n")
+			return
+		case 2:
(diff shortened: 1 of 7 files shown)


bind/gencffi_func.go
```diff
@@ -109,6 +109,36 @@ func (g *cffiGen) genFuncBody(f Func) {
 
 	g.wrapper.Printf("_cffi_helper.lib.cgo_func_%[1]s(%[2]s)\n", f.id, strings.Join(funcArgs, ", "))
 
+	if f.err {
+		switch nres {
+		case 1:
+			g.wrapper.Printf("if not _cffi_helper.lib._cgopy_ErrorIsNil(cret):\n")
+			g.wrapper.Indent()
+			g.wrapper.Printf("c_err_str = _cffi_helper.lib._cgopy_ErrorString(cret)\n")
```
sbinet: I haven't checked, but, do you know what's the memory ownership model for C-strings being passed by `_cgopy_ErrorString` ?
and what is its interaction with `ffi.string` ?

ie: who's supposed to `free` the C-string ?

corona10: @sbinet 
For about `ffi.string`. I am using this because of Py3 string handling.
It's kind of utility function, which copy the string value from string pointer and join into python string. (Internally, it uses ctypes.)

corona10: @sbinet 
Here is the reference from CFFI documentation.
>Any operation that would in C return a pointer or array or struct type gives you a fresh cdata >object. Unlike the “original” one, these fresh cdata objects don’t have ownership: they are >merely references to existing memory.

So Go will have an ownership. Hmm, it might problem when Go removes the object before Python referencing this?

corona10: It looks like I should call `cgopy_incref` and `cgopy_decref` when I referencing pointer from Go. Is that right?

bind/gengo.go
```diff
@@ -89,6 +89,11 @@ func _cgopy_ErrorString(err error) *C.char {
 	return C.CString(err.Error())
 }
 
+//export _cgopy_FreeCString
```
corona10: @sbinet 
I add this function to free the C-string manually.

sbinet: ok.

bind/gencffi_func.go
sbinet: ok.

PR #111, opened 2017-06-27
gencffi: Support bool types.

corona10:
gencffi: Support bool types.


corona10:
@sbinet PTAL

_examples/simple/simple.go
```diff
@@ -12,3 +12,10 @@ func Func() {}
 func Add(i, j int) int {
 	return i + j
 }
+
+func Bool(b bool) bool {
```
sbinet: it's a bit of a confusing test function.
why not just:

```go
func Bool(b bool) bool {
   return b
}
```

or perhaps:
```go
func Equal(a, b int) bool {
   return a == b
}
```

(or did you want to also test `bool`eans as arguments of functions? in that case I would stick with my former `Bool()` function)

corona10: > or did you want to also test booleans as arguments of functions?

Yeah, That's what I intended to do.

corona10:
Updated!

Issue #106, opened 2017-06-20
cffi: Support struct types


corona10:
Now, CFFI backend needs to generate struct types.

corona10:
Working Branch
https://github.com/corona10/gopy/tree/gsoc-usertype

corona10:
#102

corona10:
Close this issue via #113.

Issue #102, opened 2017-06-08
cffi: Implement more golang features for cffi backend.


corona10:
After https://github.com/go-python/gopy/pull/93 is landed.
We are going to focus on implementing features for the cffi backend.

- [x] Var
- [x] Const
- [x] Builtin arguments
- [x] Go maps
- [x] Go struct
- [ ] Go interfaces

rcoreilly:
for old version, closing

PR #113, opened 2017-06-30
cffi: Support struct types.

corona10:
* Generate Struct types from a CFFI engine.
* Passing a Struct type as an argument is supported.
* Passing `structs.go` test.

Update:  go-python/gopy#102
Fix: go-python/gopy#106

### DIFF
--- a/bind/gencffi.go
+++ b/bind/gencffi.go
@@ -143,8 +143,6 @@ class _cffi_helper(object):
     def cffi_cgopy_cnv_c2py_uint(c):
         return int(c)
 
-# make sure Cgo is loaded and initialized
-_cffi_helper.lib.cgo_pkg_%[1]s_init()
 `
 )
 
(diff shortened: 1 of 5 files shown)


corona10:
@sbinet

main_test.go
```diff
@@ -520,6 +520,26 @@ s2.Public = 42
...
+s1 = structs.S1{private:0}
+caught error: 'S1' object has no attribute 'private'
+s2 = structs.S2()
+s2 = structs.S2{Public:0, private:0}
+s2 = structs.S2(1)
+s2 = structs.S2{Public:1, private:0}
+caught error: S2.__init__ takes at most 1 argument(s)
+s2 = structs.S2{Public:42, private:0}
+s2.Public = 42
+caught error: 'S2' object has no attribute 'private'
```
corona10: @sbinet 
Note that error msg is changed
CPy2 version
````` python
caught error: 'structs.S2' object has no attribute 'private'

File: bind/gencffi.go [hunk 2]
```diff
@@ -169,6 +196,24 @@ func (g *cffiGen) genCffiCdef() {
 func (g *cffiGen) genWrappedPy() {
 	n := g.pkg.pkg.Name()
 	g.wrapper.Printf(cffiHelperPreamble, n)
+	g.wrapper.Indent()
+	for _, s := range g.pkg.structs {
+		g.genStructConversion(s)
+	}
+	g.wrapper.Outdent()
+
+	g.wrapper.Printf("# make sure Cgo is loaded and initialized\n")
```
sbinet: you may want to comment here why we have to put the loading of cgo here and not in the preamble.

bind/gencffi_cdef.go
```diff
@@ -5,10 +5,16 @@
 package bind
 
 import (
+	"go/types"
 	"strconv"
 	"strings"
 )
 
+func (g *cffiGen) genCdefStruct(s Struct) {
+	g.wrapper.Printf("typedef void* %s;\n", s.sym.cgoname)
```
sbinet: perhaps also generate some comment that says what this `void*` is? (ie with the Go name of the symbol being wrapped)

main_test.go
sbinet: it would be better to keep the module/package name there as well.

corona10: @sbinet 
To solve this issue, I need to change `tp_name`. Is there any way to change tp_name on Python code level?

corona10:
@sbinet Sorry for a delay. Updated!

main_test.go
sbinet: no, I don't think there is.
but it's actually probably a mistake from my end to try to sneak the package name in there.
regular python classes don't show this.
`gopy` shouldn't either.

sbinet: I'll file an issue against this.

sbinet: done: https://github.com/go-python/gopy/issues/119

sbinet:
thanks!

sbinet:
closed via 4442f0a

PR #115, opened 2017-07-02
cffi: Let Python3 handles string as ASCII.

corona10:
This is a temporary patch for Python3.
We have to update it to handle string as Unicode.
For Python2, I plan to set the default encoding with Unicode.

### DIFF
--- a/bind/gencffi.go
+++ b/bind/gencffi.go
@@ -17,7 +17,10 @@ const (
 	// discuss: https://github.com/go-python/gopy/pull/93#discussion_r119652220
 	cffiPreamble = `"""%[1]s"""
 import os
+import sys
 import cffi as _cffi_backend
+
+_PY3 = sys.version_info[0] == 3
 
 ffi = _cffi_backend.FFI()
 ffi.cdef("""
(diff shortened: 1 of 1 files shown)


corona10:
@sbinet PTAL

sbinet:
thanks.
I'll review ASAP (I am at a conference: jdev2017.fr, at the moment, so I may lag a bit...)

corona10:
np :)

bind/gencffi.go
```diff
@@ -17,8 +17,11 @@ const (
 	// discuss: https://github.com/go-python/gopy/pull/93#discussion_r119652220
 	cffiPreamble = `"""%[1]s"""
 import os
+import sys
 import cffi as _cffi_backend
 
+PY3 = sys.version_info[0] == 3
```
sbinet: perhaps better to unexport it (`s/PY3/_PY3/g`) ?

corona10:
Thanks for the review!

Updated!

sbinet:
closed with f5e31a2.

thanks.

PR #131, opened 2017-08-05
bind/py2: attach documentation for vars get/set

sbinet:


bind/gencpy.go
```diff
@@ -243,11 +243,17 @@ func (g *cpyGen) gen() error {
 
 	for _, v := range g.pkg.vars {
 		name := v.Name()
+		get := v.doc
+		set := v.doc
+		if v.doc == "" {
```
corona10: @sbinet 
Looks like this is duplicated part of this [code](https://github.com/go-python/gopy/blob/a4a47725eb8640cfbf40aca10dbf712c392b4fab/bind/gencpy.go#L319).
Is there no way to use this part?

sbinet: You're probably not wrong :)
Let me check.

sbinet: humm... well, turns out it's not completely straightforward.
`gopy`'s `cpy2` generation code is missing the concept of the step needed by the CPython2 backend to create the necessary C-structures for functions and variables (the top-level ones that are populating the C-module.)

the code you pointed out is actually completely ignoring the `Var.doc` field passed to the `Function` generation code. (because that generation code is only concerned with the wrapping/calling of the C function from CPython. the documentation is only attached during `cpyGen.gen()`.)

For now, I'd be willing to ignore this code duplication.

sbinet:
PTAL

bind/gencpy.go
corona10: @sbinet 
I agree with you. And this might be a not easy issue when we consider Python extension module structure.
How about just left `FIXME` comment.

corona10:
@sbinet
I agree with you. And this might be a not easy issue when we consider Python extension module structure.
How about just left FIXME comment.

sbinet:
now the doc is computed in one place (during `cpyGen.gen()`.)
I am fine with that.

Issue #135, opened 2017-08-09
cffi: Support map


corona10:
Working branch is here.
https://github.com/corona10/gopy/tree/support_map

corona10:
closed by #137

Issue #22, opened 2015-08-05
wrap go.maps (tp_as_sequence, tp_as_dict)


sbinet:
go.maps should be wrapped and their python counterpart should implement the sequence and dict protocols.

kellrott:
:+1: hope to see this working on soon :-)

rcoreilly:
for old version, closing

PR #137, opened 2017-08-10
cffi: Support built-in maps.

corona10:
* Support built-in maps.
* Support maps as arguments of a function.
* Support python dictionary as function parameters.
* Add 'maps' test

Fixes: #135 
Updates: #22

### DIFF
--- a/_examples/maps/maps.go
+++ b/_examples/maps/maps.go
@@ -4,13 +4,42 @@
 
 package maps
 
-func MapsFunc(t map[string]int) {
+import (
+	"sort"
+)
 
+func Sum(t map[int]float64) float64 {
+	sum := 0.0
+	for _, v := range t {
+		sum += v
+	}
+
+	return sum
 }
 
-func MapsFunc2() map[int]string {
-	return map[int]string{
-		1: "hello",
-		2: "world",
+func New() map[int]float64 {
+	return map[int]float64{
+		1: 3.0,
+		2: 5.0,
 	}
+}
+
+func Keys(t map[int]float64) []int {
+	var keys []int
+	for k, _ := range t {
+		keys = append(keys, k)
+	}
+
+	sort.Ints(keys)
+	return keys
+}
+
+func Values(t map[int]float64) []float64 {
+	var values []float64
+	for _, v := range t {
+		values = append(values, v)
+	}
+
+	sort.Float64s(values)
+	return values
 }
--- a/_examples/maps/test.py
+++ b/_examples/maps/test.py
@@ -1,0 +1,15 @@
+# Copyright 2017 The go-python Authors.  All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+from __future__ import print_function
+import maps
+
+a = maps.New()
+b = {1: 3.0, 2: 5.0}
+print('maps.Sum from Go map:', maps.Sum(a))
+print('maps.Sum from Python dictionary:', maps.Sum(b))
+print('maps.Keys from Go map:', maps.Keys(a))
+print('maps.Values from Go map:', maps.Values(a))
+print('maps.Keys from Python dictionary:', maps.Keys(b))
+print('maps.Values from Python dictionary:', maps.Values(b))
--- a/bind/gencffi_cdef.go
+++ b/bind/gencffi_cdef.go
@@ -83,6 +83,13 @@ func (g *cffiGen) genTypeCdefInit(sym *symbol) {
 		g.wrapper.Printf("extern void cgo_func_%[1]s_append(void* p0, %[2]s p1);\n", sym.id, elemType.cgoname)
 		g.wrapper.Printf("extern %[1]s cgo_func_%[2]s_item(void* p0, %[1]s p1);\n", elemType.cgoname, sym.id)
 	case sym.isMap():
+		ktyp := sym.GoType().Underlying().(*types.Map).Key()
+		etyp := sym.GoType().Underlying().(*types.Map).Elem()
+		ksym := g.pkg.syms.symtype(ktyp)
+		esym := g.pkg.syms.symtype(etyp)
+		g.wrapper.Printf("extern void cgo_func_%[1]s_set(void* p0, %[2]s p1, %[3]s p2);\n", sym.id, ksym.cgoname, esym.cgoname)
+		g.wrapper.Printf("extern %[3]s cgo_func_%[1]s_get(void* p0, %[2]s p1);\n", sym.id, ksym.cgoname, esym.cgoname)
+		g.wrapper.Printf("extern GoInt cgo_func_%[1]s_len(void* p0);\n", sym.id)
 	case sym.isSignature():
 	case sym.isInterface():
 	default:
--- a/bind/gengo.go
+++ b/bind/gengo.go
@@ -752,6 +752,7 @@ func (g *goGen) genType(sym *symbol) {
 		g.Printf("}\n\n")
 	}
 
+	g.genTypeTPMapping(sym)
 	g.genTypeTPCall(sym)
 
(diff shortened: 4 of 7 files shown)


bind/gencffi_cdef.go
```diff
@@ -83,6 +83,13 @@ func (g *cffiGen) genTypeCdefInit(sym *symbol) {
 		g.wrapper.Printf("extern void cgo_func_%[1]s_append(void* p0, %[2]s p1);\n", sym.id, elemType.cgoname)
 		g.wrapper.Printf("extern %[1]s cgo_func_%[2]s_item(void* p0, %[1]s p1);\n", elemType.cgoname, sym.id)
 	case sym.isMap():
+		ktyp := sym.GoType().Underlying().(*types.Map).Key()
+		etyp := sym.GoType().Underlying().(*types.Map).Elem()
+		ksym := g.pkg.syms.symtype(ktyp)
+		esym := g.pkg.syms.symtype(etyp)
+		g.wrapper.Printf("extern void cgo_func_%[1]s_set(void* p0, %[2]s p1, %[3]s p2);\n", sym.id, ksym.cgoname, esym.cgoname)
+		g.wrapper.Printf("extern GoString cgo_func_%[1]s_get(void* p0, %[2]s p1);\n", sym.id, ksym.cgoname)
```
sbinet: shouldn't this return `%[3]s` instead of `GoString` ? (with an additional parameter `esym.cgoname`)

_examples/maps/maps.go
```diff
@@ -4,11 +4,19 @@
...
+func MapsFunc(t map[int]string) string {
+	var buffer bytes.Buffer
+	for _, v := range t {
+		buffer.WriteString(v)
+	}
+	return buffer.String()
 }
 
-func MapsFunc2() map[int]string {
+func CreateMaps() map[int]string {
```
sbinet: as there is only one `map` being created and returned, I'd just call this `CreateMap`, or, better yet: `New`.

bind/gencffi_cdef.go
corona10: @sbinet 
Yes, you are right. It's my fault. I will update it soon.

corona10:
@sbinet PTAL.
Printing internal of map or dictionary's elements might hard because of key ordering.

sbinet:
you could have the `test.py` script sort the keys and/or elements before printing them :)

sbinet:
e.g.:

```python
print("a.keys: %s" % (sorted(a.keys()), )
```

corona10:
@sbinet 
Then we need to generate keys() and values() method for wrapped type :)
Okay, I will update it.

_examples/maps/test.py
```diff
@@ -0,0 +1,12 @@
...
+# license that can be found in the LICENSE file.
+
+from __future__ import print_function
+import maps
+
+a = maps.NewMap()
+b = {1: 3.0, 2: 5.0}
+print('maps.MapsFunc from Go map:', maps.MapsFunc(a))
+print('maps.MapsFunc from Python dictionary:', maps.MapsFunc(b))
+
```
sbinet: please add 2 more tests:

- one with a python dict whose elements are strings (and try to pass it to `maps.Func` and check we correctly detect/handle that)
- one with a Go map whose elements are also strings (and try to pass it to `maps.Func`)

bind/gengo.go
```diff
@@ -752,6 +752,97 @@ func (g *goGen) genType(sym *symbol) {
...
+			sym.id,
+			sym.cgoname,
+			ksym.cgotypename(),
+			esym.cgotypename(),
+		)
+		g.Indent()
+		g.Printf("m := (*%[1]s)(unsafe.Pointer(self))\n", sym.gofmt())
+		g.Printf("if *m == nil {\n")
+		g.Indent()
+		g.Printf("*m = make(map[%[1]s]%[2]s)\n", ksym.cgotypename(), esym.cgotypename())
```
sbinet: it's a bit un-Go-ish.
in regular Go, trying to set/modify a `nil` `map` will panic.

any reason not to do that?

bind/gengo.go
```diff
@@ -752,6 +752,97 @@ func (g *goGen) genType(sym *symbol) {
...
+		g.Outdent()
+		g.Printf("}\n")
+		if !ksym.isBasic() {
+			g.Printf("k := *(*%[1]s)(unsafe.Pointer(key))", ksym.gofmt())
+		} else {
+			if ksym.isNamed() {
+				g.Printf("k := %[1]s(key)\n", esym.gofmt())
+
```
sbinet: remove that blank empty line

bind/gengo.go
```diff
@@ -752,6 +752,97 @@ func (g *goGen) genType(sym *symbol) {
...
+				g.Printf("k := key\n")
+			}
+		}
+
+		if !esym.isBasic() {
+			g.Printf("v := *(*%[1]s)(unsafe.Pointer(value))", esym.gofmt())
+		} else {
+			if esym.isNamed() {
+				g.Printf("v := %[1]s(value)", esym.gofmt())
+
```
sbinet: remove that blank empty line

bind/gengo.go
```diff
@@ -752,6 +752,97 @@ func (g *goGen) genType(sym *symbol) {
 		g.Printf("}\n\n")
 	}
 
+	if sym.isMap() {
+		ktyp := sym.GoType().Underlying().(*types.Map).Key()
```
sbinet: this whole block (inside the `if sym.isMap()`) could be factored inside a `func (g *goGen) genTypeTPMapping(sym *symbol)` method.

so this `genType()` method would look like:

```go
g.genTypeTPMapping(sym)
g.genTypeTPCall(sym)
```

and the first line of `genTypeTPMapping` would look like:
```go
func (g *goGen) genTypeTPMapping(sym *symbol) {
    if !sym.isMap() {
        return
    }
    // ...
}
```

bind/gengo.go
```diff
@@ -752,6 +752,97 @@ func (g *goGen) genType(sym *symbol) {
...
+		)
+		g.Indent()
+		g.Printf("m := (*%[1]s)(unsafe.Pointer(self))\n", sym.gofmt())
+
+		if !ksym.isBasic() {
+			g.Printf("k := *(*%[1]s)(unsafe.Pointer(key))\n", ksym.gofmt())
+		} else {
+			if ksym.isNamed() {
+				g.Printf("k := %[1]s(key)\n", esym.gofmt())
+
```
sbinet: remove that blank empty line

bind/gengo.go
```diff
@@ -752,6 +752,97 @@ func (g *goGen) genType(sym *symbol) {
...
+			} else {
+				g.Printf("k := key\n")
+			}
+		}
+
+		g.Printf("v, _ := (*m)[k]\n")
```
sbinet: no need to use the comma-form of the map access if we aren't using its result.
just:
```go
return (*m)[k]
```

sbinet:
as a first step (generating `.keys()` and `.values()` could be hard), you could write a Go function in the `maps.go` test that does that (and let `gopy` wrap that).

bind/gengo.go
corona10: @sbinet 
Hmm, The reason why I do that is when we converted from a Python dict into the Go map.
It will call _XXXX_new() to get new map object and then it tries to insert data into a new map object from Python but it failed with 
`assignment to entry in nil map.`

corona10:
@sbinet Okay I will update it after I eat dinner 👨‍🍳  :)

bind/gengo.go
sbinet: ok.

corona10:
@sbinet PTAL

sbinet:
fixed w/ b237d1d

PR #150, opened 2018-01-01
test: Running go test without py2-cffi throws errors

rameshg87:
- There was a bug in test which didn't disable py2-cffi test backend
  even if the cffi module wasn't present on the system. This was because py2
  and py2-cffi were by default part of the enabled backends and were never disabled.
- We shouldn't have default enabled backends and should enable
  only if support is present.
- The value for testBackends needs to be only a bool instead of an int.
- Also added some log messages.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 )
 
@@ -55,12 +56,14 @@ func TestGofmt(t *testing.T) {
 	}
 }
 
-var testBackends = map[string]int{
-	"python2":  1,
-	"py2-cffi": 1,
-}
+var testBackends = map[string]bool{}
 
 func init() {
+	if os.Getenv("GOPY_TRAVIS_CI") == "1" {
+		log.Printf("Running in travis CI")
+	}
+
+	var disabled []string
 	for _, be := range []struct {
 		name   string
 		vm     string
@@ -84,11 +87,21 @@ func init() {
 		cmd.Stderr = os.Stderr
 		err := cmd.Run()
 		if err != nil {
-			log.Printf("disabling testbackend: %q", be.name)
-			continue
+			log.Printf("disabling testbackend: %q, error: '%s'", be.name, err.Error())
+			testBackends[be.name] = false
+			disabled = append(disabled, be.name)
+		} else {
+			log.Printf("enabling testbackend: %q", be.name)
+			testBackends[be.name] = true
 		}
-		log.Printf("enabling testbackend: %q", be.name)
-		testBackends[be.name] = 1
+	}
+
+	if len(disabled) > 0 {
+		log.Printf("The following test backends are not available: %s",
+			strings.Join(disabled, ", "))
+		if os.Getenv("GOPY_TRAVIS_CI") == "1" {
+			log.Fatalf("Not all backends available in travis CI")
+		}
 	}
 }
 
--- a/zsupport_test.go
+++ b/zsupport_test.go
@@ -16,8 +16,9 @@ func testPkg(t *testing.T, table pkg) {
 		backends = []string{"py2"}
 	}
 	for _, be := range backends {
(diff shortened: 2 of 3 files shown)


main_test.go
```diff
@@ -55,10 +55,7 @@ func TestGofmt(t *testing.T) {
 	}
 }
 
-var testBackends = map[string]int{
-	"python2":  1,
```
sbinet: the reasonning was to always test for/with python2 and python2+cffi as python2 is the reference (for now)

admittedly, this makes only sense when testing with travis-CI.
perhaps we should initialize `testBackends` as you do now but test (inside `init()`) if we are running with travis and forcefully enable these 2 backends back again.

main_test.go
```diff
@@ -84,11 +81,13 @@ func init() {
 		cmd.Stderr = os.Stderr
 		err := cmd.Run()
 		if err != nil {
+			log.Printf("error while checking testbackend %q: %s", be.name, err.Error())
```
sbinet: could you append the displaying of the error in the `disabling ...` printout instead?

zsupport_test.go
```diff
@@ -16,8 +17,9 @@ func testPkg(t *testing.T, table pkg) {
 		backends = []string{"py2"}
 	}
 	for _, be := range backends {
-		if _, ok := testBackends[be]; !ok {
+		if !testBackends[be] {
 			// backend not available.
+			log.Printf("Skipped testing backend %s for %s\n", be, table.path)
```
sbinet: that should be using `t.Logf` instead.

main_test.go
rameshg87: I think we can do it the opposite way. Instead of detecting travis/user, we should by default attempt to run only when all the backends are available. Otherwise, the user has to explicitly state that he/she wants to run only on the available backends using an environment variable something like GOPY_TEST_SKIP_UNAVAILABLE_BACKENDS=1. Please see next patch.

main_test.go
rameshg87: Done.

rameshg87:
Travis is happy and testing all backends now.

main_test.go
sbinet: the problem with this is we may fail (in Travis) to detect that we are not testing _e.g._ py3 because for some reason the backend isn't detected.
detecting which backends are available and run those is fine for a user.

but for travis, I think we need to be explicit and fail hard if any given backend is failing.

sbinet: so I'd be in favor of modifying the `.travis.yml` file to export/define an environment variable (say `GOPY_TRAVIS_CI=1`), detect it inside `main_test.go` and act accordingly.

what do you think?

rameshg87: Done.

zsupport_test.go
rameshg87: Done.

rameshg87:
Rebased on top of the master branch as a single commit.

rameshg87:
Can you please review and merge this pull request if it looks fine?

main_test.go
```diff
@@ -55,12 +56,14 @@ func TestGofmt(t *testing.T) {
...
-	"py2-cffi": 1,
-}
+var testBackends = map[string]bool{}
 
 func init() {
+	if os.Getenv("GOPY_TRAVIS_CI") == "1" {
+		log.Printf("Running in travis CI")
+	}
+
+	var notAvailableBackends []string
```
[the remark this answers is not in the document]
rameshg87: Done.

main_test.go
```diff
@@ -84,11 +87,21 @@ func init() {
...
+			log.Printf("enabling testbackend: %q", be.name)
+			testBackends[be.name] = true
+		}
+	}
+
+	if len(notAvailableBackends) > 0 {
+		log.Printf("The following test backends are not available: %s",
+			strings.Join(notAvailableBackends, ", "))
+		if os.Getenv("GOPY_TRAVIS_CI") == "1" {
+			panic("Not all backends available in travis CI")
```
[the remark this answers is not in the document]
rameshg87: Done.

Issue #148, opened 2017-11-30
Syntax error in generated cgo binder stub


linux4life798:
gopy fails to generate a python binding for the following Go source lib:
```go
package testlib2

import "encoding/json"

type MyStruct struct {
	AnyField json.RawMessage
}
```

The `gopy bind` command fails with the following error message:
```
/tmp/gopy-315298386/testlib2.go:252:32: expected '(', found '/'
/tmp/gopy-315298386/testlib2.go:258:32: expected '(', found '/'
2017/11/30 02:19:05 error dispatching command: exit status 1
```

I have stopped the command midway and found the syntax error in the generated Go stub.
The offending lines of generated code are the following:
```go
//export cgo_func_testlib2_encoding/json.RawMessage_MarshalJSON
func cgo_func_testlib2_encoding/json.RawMessage_MarshalJSON(self cgo_type_0x2770805872) (cgo_type_0x359372799, error) {
	res000, res001 := (*testlib2.RawMessage)(unsafe.Pointer(self)).MarshalJSON()
	return cgo_type_0x359372799(unsafe.Pointer(&res000)), res001
}

//export cgo_func_testlib2_encoding/json.RawMessage_UnmarshalJSON
func cgo_func_testlib2_encoding/json.RawMessage_UnmarshalJSON(self cgo_type_0x2770805872, arg000 cgo_type_0x359372799) (error) {
	res000 := (*testlib2.RawMessage)(unsafe.Pointer(self)).UnmarshalJSON(*(*[]byte)(unsafe.Pointer(arg000)))
	return res000
}
```

Best,

Craig

sbinet:
indeed... sorry about that.

linux4life798:
No worries! I really like the project, and I'm just curious what the fix might be.

corona10:
I will take a look this issue also. :)

sbinet:
hi,
I finally got around to have a look at this.
I fixed the original issue (`/` and `.` in symbol IDs) but there is a more fundamental issue there at work as well: right now `gopy` doesn't know how to wrap packages that use (and expose) types or identifiers from other packages.

sbinet:
see #82 as well.

rcoreilly:
this should be fixed now in new version, with `pkg` mode.

PR #154, opened 2018-01-06
gopy,bind: sanitize symbol ids

sbinet:
This CL makes sure there are no '/' nor '.' in symbol ids.
Such characters could come from a package path (e.g.: "encoding/json")
or from a fully qualified type name (e.g. json.RawMessage)

Updates go-python/gopy#148.


rcoreilly:
Closing this as no longer relevant to current version.

Issue #159, opened 2018-03-26
gopy: concurrent map writes during tests


sbinet:
Commit ab145c0 that introduced support for test coverage, also introduced a regression:

```
fatal error: concurrent map writes
goroutine 12 [running]:
runtime.throw(0x715aeb, 0x15)
	/home/travis/.gimme/versions/go1.7.6.linux.amd64/src/runtime/panic.go:566 +0x95 fp=0xc420140570 sp=0xc420140550
runtime.mapassign1(0x6c4160, 0xc420146d50, 0xc420140748, 0xc420140710)
	/home/travis/.gimme/versions/go1.7.6.linux.amd64/src/runtime/hashmap.go:553 +0x2e1 fp=0xc420140658 sp=0xc420140570
github.com/gonuts/flag.(*FlagSet).parseOne(0xc42000edc0, 0x6da101, 0x0, 0x0)
	/home/travis/gopath/src/github.com/gonuts/flag/flag.go:749 +0x345 fp=0xc420140868 sp=0xc420140658
github.com/gonuts/flag.(*FlagSet).Parse(0xc42000edc0, 0xc4200a96d0, 0x3, 0x3, 0x85d980, 0xc4200b1d40)
	/home/travis/gopath/src/github.com/gonuts/flag/flag.go:761 +0x60 fp=0xc420140898 sp=0xc420140868
github.com/gonuts/commander.(*Command).Dispatch(0xc42000ed80, 0xc4200a96d0, 0x3, 0x3, 0x1, 0x0)
	/home/travis/gopath/src/github.com/gonuts/commander/commands.go:215 +0x2ec fp=0xc420140930 sp=0xc420140898
github.com/gonuts/commander.(*Command).Dispatch(0xc42000eea0, 0xc4200a96c0, 0x4, 0x4, 0x0, 0x0)
	/home/travis/gopath/src/github.com/gonuts/commander/commands.go:187 +0x6cf fp=0xc4201409c8 sp=0xc420140930
github.com/go-python/gopy.run(0xc4200a96c0, 0x4, 0x4, 0x7120b1, 0xc)
	/home/travis/gopath/src/github.com/go-python/gopy/main.go:42 +0x9e fp=0xc420140a30 sp=0xc4201409c8
github.com/go-python/gopy.testPkgBackend(0xc420074540, 0x70ef62, 0x3, 0x70ffaf, 0x7, 0x7120b1, 0xc, 0xc420010b00, 0x1, 0x1, ...)
[...]

PR #162, opened 2018-03-26
gopy: use mutex to protect against concurrent map writes during tests

initialed85:
Apologies if some of the process is wrong- this is the first time I've submitted a PR to another repo on GitHub (we use Bitbucket and Mercurial at work).

This PR aims to address https://github.com/go-python/gopy/issues/159 wherein the tests will not pass when run in parallel due to concurrency issues.

All tests pass on my machine when invoked with `GODEBUG=cgocheck=0 go test -a -v -parallel=16 -run=`

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
+	"sync"
 	"testing"
 )
 
@@ -57,6 +58,8 @@ func TestGofmt(t *testing.T) {
 }
 
 var testBackends = map[string]bool{}
+
+var tableMutex sync.Mutex
 
 func init() {
 	if os.Getenv("GOPY_TRAVIS_CI") == "1" {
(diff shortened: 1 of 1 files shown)


main_test.go
```diff
@@ -112,6 +115,16 @@ type pkg struct {
...
 func testPkgBackend(t *testing.T, lang, pycmd string, table pkg) {
+	tableMutex.Lock()
+	tempTable := pkg{
+		path: table.path,
+		lang: table.lang,
+		want: table.want,
+	}
+	tableMutex.Unlock()
+
+	table = tempTable
```
sbinet: I don't understand why copying a value that contains slices of stuff would cure the "concurrent map writes" error.
could you explain this ?
also, to be sure this cures that problem, you should re-add the 2 commits that introduced that feature :)

(but thanks for taking a stab at this!)

initialed85: Hmm, my apologies- perhaps I've fixed something that wasn't broken at the same time as those commits were reverted or something; I did have failing tests that stopped failing after doing this, I'm almost sure of it.

Let me dig deeper and get back to you- may as well can the PR, I daresay it's the barking up the wrong tree altogether!

initialed85:
Cancelling PR

PR #163, opened 2018-03-26
gopy: fix concurrent map writes during tests

initialed85:
Second attempt at fixing https://github.com/go-python/gopy/issues/159; I've included the two commits that were reverted.

Tests pass on my system under Go1.9 and Go1.10.


sbinet:
thanks!

could you separate into 3 commits ?
the first 2 commits being the ones that have been reverted (and back again)
and then the one that fixes the issue ?

that makes for a clearer history of what fixes what.

thanks again.

initialed85:
@sbinet Okay, I think that's split up in the way you're after

bind/package.go
```diff
@@ -28,9 +29,17 @@ type Package struct {
 	funcs   []Func
 }
 
+var (
+	mu sync.Mutex
```
sbinet: could you move the declaration of that mutex closer to the declaration of the variable it's guarding access to?
(ie: just above `universe` in the file `bind/symtab.go`)

bind/package.go
```diff
@@ -28,9 +29,17 @@ type Package struct {
...
+)
+
 // NewPackage creates a new Package, tying types.Package and ast.Package together.
 func NewPackage(pkg *types.Package, doc *doc.Package) (*Package, error) {
+	// protection for parallel tests
+	mu.Lock()
 	universe.pkg = pkg // FIXME(sbinet)
+	mu.Unlock()
```
sbinet: hum... wouldn't this introduce some inconsistency?

if `test1` enters here first, the `pkg` field points to `pkg1`.
then, the lock is released.
but `test2` could enter, grab the lock, put `pkg2` in the field *before* `test1` finishes to do its thing: `test1` might access `universe.pkg` while `pkg` is in fact `pkg2`.

I'd probably `defer` the call to `mu.Unlock()` to make sure the processing of the `Package` is consistent.

bind/package.go
initialed85: Will do

bind/package.go
initialed85: That's a fair point- I was really only thinking of protecting the object rather than the impact overall.

Running the tests with a deferred `.Unlock()` encompassing all of `NewPackage()` doesn't seem to take much longer, so that's good. Will update the PR

corona10:
@sbinet @initialed85
Is this matrix generater is really needed?Since some of matrixes are not accurated, I don't know this generater is really needed.

For example, test hi is not actually failed test for cffi. Becauae Cffi and C API has a different mechhanism for buffer interface.

corona10:
Or we need to fix some tests.

corona10:
I am agree with this Cl that it would help to people can know easily about the supported status of each API engine.

But I am just worry about that people can misunderstand the status of each API since some of tests are not equivalent environment for all API. One of the example is hi.

initialed85:
@corona10 I can't really weigh in on that one I'm afraid- my PR mostly addresses concurrency issues that came about from https://github.com/go-python/gopy/pull/163/commits/0a077a485afbee062bdcee36e1a3bd1c8d59573e.

https://github.com/go-python/gopy/pull/163/commits/fa9706008614ce3b9d612a5b575dc619fc13f79a was related to that issue and both were reverted, so they were reintroduced with this PR.

corona10:
@initialed85 @sbinet 
After merging this PR, I suggest that we should update some tests based on supporting features, not behavior based tests :)

corona10:
@initialed85 
Can you rebase your PR? 
Since I added a test to my last PR(https://github.com/go-python/gopy/pull/161),
there are some conflicts on main_test.go.

bind/package.go
```diff
@@ -28,9 +28,15 @@ type Package struct {
 	funcs   []Func
 }
 
+var ()
```
sbinet: this could be removed completely :)

sbinet:
this feature matrix could indeed be rethought and reworked to be more feature oriented rather than name-of-the-test oriented :)

being able to extract a code coverage from the tests is, OTOH, quite useful.
(we just need to actually extract it and display it...)

initialed85:
@sbinet removed unused `var ()` as suggested- cheers

sbinet:
needs https://github.com/go-python/license/pull/9

PR #164, opened 2018-06-11
Add the ability to change method and function names

ondra:
We're porting a library to Go using Gopy and we need to keep the method and function names the same as in the old implementation.

Currently, only exported functions, starting with an uppercase character, can have a binding generated.

This PR adds a directive, which allows changing the name in the following way:

```
//python:name test
func Test() {
   ...
}
```

### DIFF
--- a/_examples/rename/rename.go
+++ b/_examples/rename/rename.go
@@ -1,0 +1,19 @@
+// Copyright 2018 The go-python Authors.  All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// package rename tests changing the names of methods and functions
+package rename
+
+//gopy:name say_hi
+func SayHi() string {
+	return "hi"
+}
+
+type MyStruct struct {
+}
+
+//gopy:name say_something
+func (s *MyStruct) SaySomething() (something string) {
+	return "something"
+}
--- a/_examples/rename/test.py
+++ b/_examples/rename/test.py
@@ -1,0 +1,11 @@
+# Copyright 2018 The go-python Authors.  All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+## py2/py3 compat
+from __future__ import print_function
+
+import rename
+
+print(rename.say_hi())
+print(rename.MyStruct().say_something())
--- a/bind/utils.go
+++ b/bind/utils.go
@@ -158,3 +158,20 @@ func getGoVersion(version string) (int64, int64, error) {
 	minor, _ := strconv.ParseInt(version_info[1], 10, 0)
 	return major, minor, nil
 }
+
+func extractPythonName(gname, gdoc string) (string, string, error) {
+	const PythonName = "\ngopy:name "
+	i := strings.Index(gdoc, PythonName)
+	if i < 0 {
+		return gname, gdoc, nil
+	}
+	s := gdoc[i+len(PythonName):]
+	if end := strings.Index(s, "\n"); end > 0 {
+		validIdPattern := regexp.MustCompile(`^[\pL][\pL_\pN]+$`)
+		if !validIdPattern.MatchString(s[:end]) {
+			return "", "", fmt.Errorf("gopy: invalid identifier: %s", s[:end])
+		}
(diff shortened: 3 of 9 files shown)


sbinet:
nice!
the test failure is just because you didn't run `gofmt` on the modified code. could you please `gofmt` your code? (thanks!)

to mimic what `cgo` does, perhaps we could instead name that directive like so:
```go
//gopy:linkname Test test
func Test() { ... }
```

_ie_ `//gopy:linkname <GoName> <target_python_name>`

what do you think?

sbinet:
(also, could you please add a test for this new feature?)

ondra:
Your proposed syntax looks fine to me, but I'm not sure about providing `GoName` explicitly.

Currently the directive needs to be on the line directly preceding the function, shall we keep it like that and check whether this is the right function or allow the directive anywhere in the source?

I don't have a strong opinion either way.

ondra:
There is now an example in `_examples/rename/` and a test.

I changed the directive to `//gopy:name ...` instead of `//python:name ...`, but I think it is better to keep the semantics this way, because the code is simple and orthogonal to the rest of the implementation.

Should `<GoName>` be required to be specified explicitly, I, as a user of gopy, would expect the binding generation to fail if the name is specified in the wrong place, for example when a space is inserted in between the function signature. This means checking every comment to give meaningful results.

I can do it, but I think it means that the implementation will be either too complex or slightly broken anyway.

What do you think?

sbinet:
I have been trying to find out why `cgo` and `Go` did it that way.
I suppose this was because at the time we didn't have the nice `go/build` and `go/doc` packages.

I am tempted to go with `//gopy:name py_name`.
I'll review the PR more in depth shortly :)

(thanks again!)

_examples/rename/rename.go
```diff
@@ -0,0 +1,19 @@
...
+
+// package rename tests changing the names of methods and functions
+package rename
+
+//gopy:name say_hi
+func SayHi() string {
+	return "hi"
+}
+
+type MyStruct struct {
```
sbinet: couldn't types be renamed as well?

_examples/rename/test.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2015 The go-python Authors.  All rights reserved.
```
sbinet: `s/2015/2018/`

bind/utils.go
```diff
@@ -158,3 +158,16 @@ func getGoVersion(version string) (int64, int64, error) {
 	minor, _ := strconv.ParseInt(version_info[1], 10, 0)
 	return major, minor, nil
 }
+
+func extractPythonName(gname, gdoc string) (string, string) {
```
sbinet: could you add a few table-based tests for this function?
something like:
```go
funcTestExtractPythonName(t*testing.T) {
    for _, tc := range []struct{ doc, wantName, wantDoc string } { ... } {
         ...
    }
}
```

also, it would be great to present the user with a nice error message (instead of some (cryptic) compilation error?) when there's a syntax error in the directive. like:
```go
//gopy:name invalid name_with_a_space
func Foo() { ... }
```

ondra:
I think repeating the name in the directive makes sense when no code is emitted by default as an additional check, in other cases it's trouble.

bind/utils.go
ondra: I added a simple check. A more complete one should test that the identifier is valid C/Go/Python, but this should be done for all emitted strings in general, I think. (One day:)

_examples/rename/test.py
ondra: fixed

_examples/rename/rename.go
ondra: It would involve a lot more changes, at the very least renaming would be necessary for generating method receivers/arguments. I'd say that it's not worth the trouble at this point.

bind/utils.go
ondra: The code still just panics on error. But perhaps it keeps the code shorter and more readable than propagating the errors all the way up.

_examples/rename/rename.go
sbinet: fair enough :)

sbinet:
thanks again!

PR #165, opened 2018-06-11
Support for pointer-to-slice parameters

ondra:
Implement support for pointer-to-slice parameters for methods and functions:

Python:
```
import tst
s = tst.StrSlice()
tst.Fill(s)
print s
```

Go:
```
package tst
type StrSlice []string
func Fill(s *StrSlice) {
        *s = StrSlice{"A", "B"}
}
```

This doesn't implement pointer support in general (as we didn't need it), but the fixes should make the implementation easier.

### DIFF
--- a/bind/gencpy_type.go
+++ b/bind/gencpy_type.go
@@ -20,6 +20,9 @@ func (g *cpyGen) genType(sym *symbol) {
 	if sym.isBasic() && !sym.isNamed() {
 		return
 	}
+	if sym.isPointer() {
+		return
+	}
 
 	g.decl.Printf("\n/* --- decls for type %v --- */\n", sym.gofmt())
 	if sym.isBasic() {
--- a/bind/gengo.go
+++ b/bind/gengo.go
@@ -277,26 +277,7 @@ func (g *goGen) genFuncBody(f Func) {
 	}
 
 	g.Printf("%s.%s(", g.pkg.Name(), f.GoName())
-
-	args := sig.Params()
-	for i, arg := range args {
-		tail := ""
-		if i+1 < len(args) {
-			tail = ", "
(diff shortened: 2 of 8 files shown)


ondra:
I'd like to support the other backends and py3k eventually, but this is what we're using now.

bind/gengo.go
```diff
@@ -588,6 +603,10 @@ func (g *goGen) genType(sym *symbol) {
 	if sym.isBasic() && !sym.isNamed() {
 		return
 	}
+	if (sym.kind & skPointer) != 0 {
```
sbinet: please remove this debugging leftover :)

bind/gencpy_type.go
```diff
@@ -20,6 +20,9 @@ func (g *cpyGen) genType(sym *symbol) {
 	if sym.isBasic() && !sym.isNamed() {
 		return
 	}
+	if (sym.kind & skPointer) != 0 {
```
sbinet: couldn't this be rewritten as:
```go
if sym.isPointer() {
    return
}
```

ondra:
How would you want the test to look like? Is it enough to do it through `_examples`?

I've been digging through the code trying to make pointers work better in general, many changes would be necessary, but I'm missing the bigger picture, so I don't think I can fix it now.

Are you willing to merge this without larger changes? I do _think_ the code is slightly better than before.

sbinet:
Yes, testing via the usual _examples is what I had in mind.
And I agree, the code looks better.
Also yes: willing to merge this once there's a test :)

ondra:
Great:)! The test is there.
I'll add support for CFFI later.
Thanks for the review, and for GoPy, of course.

PR #170, opened 2018-11-27
gopy/bind: encode strings passed to Go using the CFFI backend as UTF-8

ondra:
Strings passed the other way are already UTF-8 encoded, this makes everything symmetric and working in many more cases.

### DIFF
--- a/bind/gencpy.go
+++ b/bind/gencpy.go
@@ -103,11 +103,25 @@ cgopy_cnv_c2py_bool(GoUint8 *addr) {
 
 static int
 cgopy_cnv_py2c_string(PyObject *o, GoString *addr) {
-	const char *str = PyString_AsString(o);
-	if (str == NULL) {
-		return 0;
+	if (PyUnicode_Check(o)) {
+		o = PyUnicode_AsUTF8String(o);
+		if (o == NULL) {
+			return 0;
+		}
+		const char *str = PyString_AsString(o);
+		if (str == NULL) {
+			Py_DECREF(o);
+			return 0;
+		}
+		*addr = _cgopy_GoString((char*)str);
+		Py_DECREF(o);
+	} else {
+		const char *str = PyString_AsString(o);
+		if (str == NULL) {
+			return 0;
+		}
+		*addr = _cgopy_GoString((char*)str);
 	}
-	*addr = _cgopy_GoString((char*)str);
 	return 1;
 }
 
--- a/main_test.go
+++ b/main_test.go
@@ -213,6 +213,7 @@ var features = map[string][]string{
 	"_examples/maps":      []string{"py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
 	"_examples/gostrings": []string{"py2", "py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
 	"_examples/rename":    []string{"py2", "py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
+	"_examples/unicode":   []string{"py2", "py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
 }
 
 func TestHi(t *testing.T) {
@@ -873,6 +874,19 @@ func TestSlicePtr(t *testing.T) {
 		want: []byte(`sliceptr.IntVector{1, 2, 3}
 sliceptr.IntVector{1, 2, 3, 4}
 sliceptr.StrVector{"1", "2", "3", "4"}
+`),
+	})
+}
+
+func TestUnicode(t *testing.T) {
+	t.Parallel()
+	path := "_examples/unicode"
+	testPkg(t, pkg{
+		path: path,
+		lang: []string{"py2", "py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
+		want: []byte(`encoding.HandleString(bytestr) -> Python byte string
+encoding.HandleString(unicodestr) -> Python Unicode string 🐱
+encoding.GetString() -> Go Unicode string 🐱
 `),
 	})
 }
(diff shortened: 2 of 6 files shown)


corona10:
@ondra
Thanks for handle this issue.
But in this case, if a user uses cffi with Python2, does it handling string as UTF-8 on your PR?

ondra:
@corona10, the behavior Python 2 behavior is not changed in this PR.

I've been thinking about string handling some more and I believe that it deserves a more thorough treatment.

Currently, generated CPython extensions only accept `str` (bytestrings) function arguments, but fail with  `unicode` arguments. CFFI extensions do the same in Python 2, but in Python 3, only `unicode` typed arguments containing ASCII characters work, bytestrings do not. This PR makes Python 3 CFFI bindings support general `unicode` arguments.

Ideally, I think that ideally:
-  `[]byte` from Go would be exposed in Python as Py2 `str` or the equivalent Py3 `bytes` only
- `string` from Go would be exposed in Python as Py2 `unicode` or the equivalent Py3 `str` only

But this is not backwards compatible with the current behavior and it is perhaps too strict, at least with respect to Python 2 standards.

So, the behavior I'd like to implement is:
- pass Py2 `str` and Py3 `bytes` to Go directly
- pass Py2 `unicode` and Py3 `str` encoded as UTF-8

What do you think?

ondra:
The updated commit implements support for bytestrings and Unicode strings in Python 2 and 3.

corona10:
@ondra 
The reason I only supported ASCII was to make align with CPython2 extension.
So this PR should update CPython2 extension generator

I think that we need @sbinet 's opinion.

sbinet:
I think that now - December 2018 - Python-2 shouldn't be something that ties us up.
we should try to generate the best, idomatic, wrapping code for Python-3.

_ie:_ I wouldn't be against a limited breakage wrt Python-2.
but, for Python-3, we need to have consistent code+semantics being generated for `cffi` and `cpython3`.

ondra:
To summarize:

Current behavior:

| | bytes | unicode |
|-|--------------|-------------|
|Python 2 CPython | pass directly | encode using the default encoding |
|Python 2 CFFI | pass directly | error |
|Python 3 CFFI | error | encode as ASCII |

Proposed behavior implemented in this PR:

| | bytes | unicode |
|-|--------------|-------------|
|Python 2 CPython | pass directly | encode as UTF-8 |
|Python 2 CFFI | pass directly | encode as UTF-8 |
|Python 3 CFFI | pass directly | encode as UTF-8 |

I'm inclined to conclude that even though the behavior in Python 2 is not the same, the new behavior is much more desirable.

At least in Python 3 the generated code _could_ be more strict, but I don't think this is desirable -- it would be necessary to either behave differently in Python 2 and 3 or break the current Python 2 behavior.

SWIG-generated modules accept both bytes and unicode in Python 2 and 3. I think that in an interface between different languages this is a good thing, as a binding generator exists to reduce the impedance mismatch between languages.

ondra:
Also, I'd like to know what do you think about string return values.

Currently, the result is always `str` and therefore differing between Python 2 (==`bytes`) and 3 (==`unicode`).

I'd like the behavior to be consistent, but I also strongly prefer the functions to return `unicode`.

As Python 2 `str` and `unicode` are almost always interchangeable (they inherit from `basestring`) and Go strings map naturally to `unicode`, I think that it would be possible and perhaps desirable to return unicode even in Python 2. What do you think? It could also be configurable per function, globally or at runtime, but I imagine that this would not be straightforward to implement cleanly.

main_test.go
```diff
@@ -875,6 +876,19 @@ sliceptr.StrVector{"1", "2", "3", "4"}
...
 
+func TestUnicode(t *testing.T) {
+	t.Parallel()
+	path := "_examples/unicode"
+	testPkg(t, pkg{
+		path: path,
+		lang: []string{"py2", "py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
+		want: []byte(`encoding.HandleString(bytestr) -> Python byte string
+encoding.HandleString(unicodestr) -> Python Unicode string 🐱
+encoding.GetString() -> Go Unicode string 🐱
```
corona10: Please remove emoji

main_test.go
```diff
@@ -875,6 +876,19 @@ sliceptr.StrVector{"1", "2", "3", "4"}
...
 }
 
+func TestUnicode(t *testing.T) {
+	t.Parallel()
+	path := "_examples/unicode"
+	testPkg(t, pkg{
+		path: path,
+		lang: []string{"py2", "py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
+		want: []byte(`encoding.HandleString(bytestr) -> Python byte string
+encoding.HandleString(unicodestr) -> Python Unicode string 🐱
```
corona10: Please remove emoji

sbinet: I assume the emoji is there to test utf-8 :)

corona10: Ah.. you are right.

sbinet:
@ondra I agree with your proposed behaviour in https://github.com/go-python/gopy/pull/170#issuecomment-443800989.

I am also enclined to agree w/ your comment in https://github.com/go-python/gopy/pull/170#issuecomment-443805729.
but perhaps this could be part of a later PR (instead of lumping it in there) ?

ondra:
@sbinet Yes, the return values are only tangentially related, I'll open a new PR later.

bind/gencpy.go
```diff
@@ -103,11 +103,24 @@ cgopy_cnv_c2py_bool(GoUint8 *addr) {
...
-	if (str == NULL) {
-		return 0;
+	if (PyUnicode_Check(o)) {
+		o = PyUnicode_AsUTF8String(o);
+		if (o == NULL) {
+			return 0;
+		}
+		const char *str = PyString_AsString(o);
+		if (str == NULL) {
+			return 0;
```
sbinet: add `Py_DECREF(o)` or `Py_XDECREF(o)` ? (otherwise, it seems to me we leak it.)

ondra: There is `Py_DECREF(o)` on line 116.

Can `o` be released sooner? `str` points into `o`.

sbinet:
closed via https://github.com/go-python/gopy/pull/173

PR #173, opened 2018-12-13
Pass Go strings to Python as Unicode objects

ondra:
This PR builds on https://github.com/go-python/gopy/pull/170.

Currently, Python 3 CFFI returns Unicode, but Python 2 backends do not.

In Python 2, both byte and Unicode strings inherit from basestring and are interchangeable for most purposes. Some code might break, as Python 2 tries to encode Unicode strings to PYTHONIOENCODING while doing IO, and the specific encoding might be unable to represent the characters present in the string. Programs which ignore encoding _and_ have the need to output fancy characters suffer.

CPython 2 docstrings as Unicode are not implemented by this PR, but I do not think this is an issue.

Making the tests portable is difficult -- only the test for Unicode needs to handle IO encoding seriously. In the others I just strip the potential `u` which `repr` uses to mark Unicode strings in Python 2.

### DIFF
--- a/_examples/unicode/test.py
+++ b/_examples/unicode/test.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+
+# Copyright 2018 The go-python Authors.  All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+## py2/py3 compat
+from __future__ import print_function, unicode_literals
+
+import encoding
+
+bytestr = b"Python byte string"
+unicodestr = u"Python Unicode string 🐱"
+
+bytestr_ret = encoding.HandleString(bytestr)
+unicodestr_ret = encoding.HandleString(unicodestr)
+
+print("encoding.HandleString(bytestr) ->", bytestr_ret)
+print("encoding.HandleString(unicodestr) ->", unicodestr_ret)
+
+gostring_ret = encoding.GetString()
+print("encoding.GetString() ->", gostring_ret)
+
--- a/_examples/unicode/test.py
+++ b/_examples/unicode/test.py
@@ -7,7 +7,16 @@
 ## py2/py3 compat
 from __future__ import print_function, unicode_literals
 
+import sys
+
 import encoding
+
+# There is no portable way of outputting Unicode in Python 2/3 -- we encode
+# them manually and sidestep the builtin encoding machinery
+try:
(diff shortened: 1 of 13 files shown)


bind/gencffi.go
```diff
@@ -87,8 +88,8 @@ class _cffi_helper(object):
 
     @staticmethod
     def cffi_cgopy_cnv_py2c_string(o):
-        if _PY3:
-            o = o.encode('ascii')
+        if _PY3 and isinstance(o, str) or not _PY3 and isinstance(o, unicode):
```
sbinet: could you put explicit parentheses?
with the mix of `and` and `or`, it's a bit hard to understand exactly what this is doing (well, of course "everybody" knows the precedence of `or` and `and`, but...)

ondra: Good idea -- it also avoids having to think about the weird operator chaning semantics in Python.

_examples/unicode/test.py
```diff
@@ -7,17 +7,30 @@
 ## py2/py3 compat
 from __future__ import print_function, unicode_literals
 
+import sys
+
 import encoding
 
+# There is no portable way of outputting Unicode in Python 2/3 -- we encode
+# them manually and sidestep the builtin encoding machinery
+try: binary_stdout = sys.stdout.buffer  # Python 3
```
sbinet: cosmetics: could you align `binary_stdout` with the one from the line below ?

bind/gencpy.go
```diff
@@ -297,10 +297,16 @@ func (g *cpyGen) gen() error {
...
-		g.pkg.pkg.Name(),
-		g.pkg.doc.Doc,
-	)
+	if g.pkg.doc.Doc != "" {
+		g.impl.Printf("module = Py_InitModule3(%[1]q, cpy_%[1]s_methods, %[2]q);\n\n",
+			g.pkg.pkg.Name(),
+			g.pkg.doc.Doc,
+		)
+	} else {
+		g.impl.Printf("module = Py_InitModule3(%[1]q, cpy_%[1]s_methods, 0);\n\n",
```
sbinet: `s/0/NULL/` ?

_examples/unicode/test.py
ondra: I put the blocks on separate lines as sane people and PEP8 recommend.

bind/gencpy.go
ondra: ok

PR #176, opened 2019-01-14
gopy/bind: enable subclassing of generated CPython objects

ondra:
To mimic a SWIG library I'm porting, I'd like to inherit from GoPy-generated objects.

I believe that switching the flag should be everything that is necessary for this to work. GoPy objects do not serve as collections of Python objects, so handling GC is not needed and the lifecycle seems to be handled well enough so that no resources are leaked.

### DIFF
--- a/_examples/structs/test.py
+++ b/_examples/structs/test.py
@@ -53,3 +53,21 @@ try:
 except Exception as err:
     print("caught error: %s" % (err,))
     pass
+
+class S2Child(structs.S2):
+    def __init__(self, a, b):
+        super(S2Child, self).__init__(a)
+        self.local = b
+    def __str__(self):
+        return ("S2Child{S2: %s, local: %d}"
+                % (super(S2Child, self).__str__(), self.local))
+
+try:
+    s2child = S2Child(42, 123)
+    print("s2child = %s" % (s2child,))
+    print("s2child.Public = %s" % (s2child.Public,))
+    print("s2child.local = %s" % (s2child.local,))
+    print("s2child.private = %s" % (s2child.private,))
+except Exception as err:
+    print("caught error: %s" % (err,))
+    pass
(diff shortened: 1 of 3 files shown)


_examples/structs/test.py
```diff
@@ -53,3 +53,18 @@
 except Exception as err:
     print("caught error: %s" % (err,))
     pass
+
+class S2_descendant(structs.S2):
```
sbinet: what about `s/S2_descendant/S2Kid/` ?

_examples/structs/test.py
```diff
@@ -53,3 +53,18 @@
...
+    def __init__(self, n):
+        super(S2_descendant, self).__init__(n)
+        self.local = n
+
+try:
+    s2_descendant = S2_descendant(43)
+    print("s2_descendant = %s" % (s2_descendant,))
```
sbinet: what about implementing `__str__` on `S2Kid` so the printout makes more sense (and actually test "polymorphism") ?

_examples/structs/test.py
```diff
@@ -53,3 +53,18 @@
...
     print("caught error: %s" % (err,))
     pass
+
+class S2_descendant(structs.S2):
+    def __init__(self, n):
+        super(S2_descendant, self).__init__(n)
+        self.local = n
+
+try:
+    s2_descendant = S2_descendant(43)
```
sbinet: `s/s2_descendant/kid/`

_examples/structs/test.py
ondra: Good idea.

_examples/structs/test.py
ondra: I renamed it to S2Child, so that it is more in line with standard terminology.

sbinet:
still seeing `S2_descendant` :)

ondra:
sorry, I forgot to actually commit;)

sbinet:
hum... it seems your latest commit also captured unrelated changes (pertaining to pointers-to-slices.)
could we stay on topic? :)

sbinet:
also: could you rebase on top of latest `master` (d30e0e6 as of now) so the Windows CI is happy?

ondra:
sorry, I pushed that in a hurry:/

https://github.com/go-python/gopy/pull/170, https://github.com/go-python/gopy/pull/173 and this PR are now rebased on tip.

sbinet:
no worries.

PR #179, opened 2019-01-21
bind: fix CFFI sequence indexing

ondra:
Indices should be `int`s instead of the element type.


sbinet:
I think that's https://github.com/go-python/gopy/commit/179d00430500d6244a94b92b8ca374930eff6303, right?

ondra:
Indeed!

(I missed some cases w.r.t. the Unicode string handling, I'll make a PR later.)

PR #180, opened 2019-03-07
major rewrite to work with current Go versions

rcoreilly:
This is a major rewrite, using the idea of int64 handles instead of passing pointers around in cgo / python.  It all works, and along the way lots of additional functionality added to be able to handle full multi-package systems, and even generate a standalone python executable as was needed for my GoGi gui system where the main thread is taken up by the gui event loop.

tests all pass for py3 but py2 has some different output sequencing so produces diff outputs sometimes.

I also switched over to using pybindgen for speed on cpython (and ease of use) -- cffi should be possible and not too hard (hopefully just changing the generating of the build.py script).  It handles all the lower-level glue so the cgo wrappers and .py wrappers all look very straightforward.  Switched to more go-standard camel-case, shorter naming style instead of all the kebab-case cgo names that I found hard to parse.. generated code seems pretty readable to me now.

.py wrappers support full container features on maps and slices.  constants are just directly declared in .py wrapper -- no going through go.  arrays are treated as go-generated only given their fixed size -- use slices for any py-side created lists.

overall, the translation for a given .go file to python do the same thing with the wrapped code is remarkably similar and illustrates the great fit between the two languages -- go is a nice subset of python and using go for fast backend computation makes a lot of sense.  see the GoGi demo case linked on the updated README.md for an example.


sbinet:
Nice!
It will take me some time to digest it (and I am on the road till the end of the week) but I'll get to it!

@corona10 do you want to take at stab at this?

corona10:
Nice! Sure, I will take a look at this PR by this weekend.

_examples/hi/hi.go
```diff
@@ -113,6 +122,35 @@ func (p *Person) Salary(h int) (int, error) {
...
+
+func (p *Person) GetAge() int {
+	return p.Age
+}
+
+func (p *Person) SetName(nm string) {
+	p.Name = nm
+}
+
+func (p *Person) SetAge(ag int) {
```
sbinet: `s/ag/age/`

_examples/hi/hi.go
```diff
@@ -113,6 +122,35 @@ func (p *Person) Salary(h int) (int, error) {
...
 
+func (p *Person) GetName() string {
+	return p.Name
+}
+
+func (p *Person) GetAge() int {
+	return p.Age
+}
+
+func (p *Person) SetName(nm string) {
```
sbinet: `s/nm/n/` (or `name`)

_examples/hi/hi.go
```diff
@@ -113,6 +122,35 @@ func (p *Person) Salary(h int) (int, error) {
...
+	p.Name = nm
+}
+
+func (p *Person) SetAge(ag int) {
+	p.Age = ag
+}
+
+func (p *Person) SetFmS2(s2arg structs.S2) {
```
sbinet: `s/s2arg/s2/`

_examples/hi/hi.go
```diff
@@ -113,6 +122,35 @@ func (p *Person) Salary(h int) (int, error) {
...
+	p.Age = s2arg.Public
+}
+
+func (p *Person) SetFmS2Ptr(s2arg *structs.S2) {
```
sbinet: `s/s2arg/s2/`

_examples/hi/hi.go
```diff
@@ -139,3 +177,21 @@ type Floats []Float
 
 // Eval evals float64
 type Eval func(f float64) float64
+
+// PersIface is an interface into the person type
```
sbinet: `s/person type/person type./`

_examples/hi/test.py
```diff
@@ -167,8 +167,8 @@
 
 ## test Couple.__init__
 print("--- Couple.__init__")
-c = hi.Couple(hi.Person("p1", 42))
-print(c)
+#c = hi.Couple(hi.Person("p1", 42))
```
sbinet: why doesn't this work anymore?
(is there a FIXME somewhere?)

_examples/iface/iface.go
```diff
@@ -27,3 +30,13 @@ func CallIface(v Iface) {
 	v.F()
 	cpkg.Printf("iface.CallIface... [DONE]\n")
 }
+
+// by default, interface{} is converted to string (most universal type)
+func IfaceString(str interface{}) {
+	cpkg.Printf("iface as string: %v\n", str)
+}
+
+//gopy:interface=handle this magic directive says, treat it as a handle
```
sbinet: could/shouldn't the directive be put on its own line (to better follow all the other `//go:something` directives) ?

_examples/maps/test.py
```diff
@@ -7,11 +7,47 @@
 
 a = maps.New()
 b = {1: 3.0, 2: 5.0}
+
+# set to true to test all the builtin map functionality
+# due to the random nature of map access in Go, this will
+# cause go test to fail randomly so it is off by default
+# but should be re-enabled when anything significant is
+# changed!
+testall = False
```
sbinet: we could sort the lists returned by `keys()` and `values()`.
so we don't have to enable/disable this.

_examples/named/named.go
```diff
@@ -5,32 +5,38 @@
...
-func (x XX) Value() float32 { return float32(x) }
-
-// Value returns a float32 value
-func (x XXX) Value() float32 { return float32(x) }
-
-// Value returns a float32 value
-func (x XXXX) Value() float32 { return float32(x) }
-
-type Str string
+// todo: not dealing with named basic types -- not very intuitive to use
```
sbinet: `s/todo/TODO/`

_examples/unicode/test.py
```diff
@@ -21,12 +21,13 @@
 bytestr = b"Python byte string"
 unicodestr = u"Python Unicode string 🐱"
 
-bytestr_ret = encoding.HandleString(bytestr)
+# todo: need conversion from bytestr to string -- not sure pybindgen can do it?
```
sbinet: `s/todo/TODO/`

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+	"strings"
+)
+
+// this version uses pybindgen and a generated .go file to do the binding
+
+// for all preambles: 1 = name of package (outname), 2 = cmdstr
+
+// GoHandle is the type to use for the Handle map key, go-side
+// could be a string for more informative but slower handles
+var GoHandle = "int64"
```
sbinet: please use a block `var`.
actually, couldn't we use `const` instead?

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+static inline void gopy_err_handle() {
+	if(PyErr_Occurred() != NULL) {
+		PyErr_Print();
+	}
+}
+%[8]s
+*/
+import "C"
+import (
+	"github.com/goki/gopy/gopyh" // handler
```
sbinet: this would need to be `go-python/gopy` right?

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+
+	goExePreambleGo = `
+// wchar version of startup args
+var wargs []*C.wchar_t
+
+//export GoPyMainRun
+func GoPyMainRun() {
+	// need to encode char* into wchar_t*
+	for i := range os.Args {
+		wargs = append(wargs, C.Py_DecodeLocale(C.CString(os.Args[i]), nil))
```
sbinet: what's the lifetime of the c-string created with `C.CString` here?
do we need to `C.free` it?
this probably needs a comment.

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+# packages imported by this package listed below:
+
+%[6]s
+
+`
+
+	GoPkgDefs = `
+import collections
+	
+class GoClass(object):
```
sbinet: couldn't this be just `Class`?
the name of the module being `go`...

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+	# note: pip install pybindgen to get pybindgen if this fails
+	$(PYTHON) build.py
+	# build the executable
+	- rm %[1]s_go$(LIBEXT)
+	$(GOBUILD) -o py%[1]s
+	
+`
+)
+
+// actually, the weird thing comes from adding symbols to the output, even with dylib!
```
sbinet: remove?

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+	MakefileTemplate = `# Makefile for python interface for package %[1]s.
+# File is generated by gopy. Do not edit.
+# %[2]s
+
+GOCMD=go
+GOBUILD=$(GOCMD) build
+GOIMPORTS=goimports
+PYTHON=%[4]s
+PYTHON_CFG=$(PYTHON)-config
+GCC=gcc
```
sbinet: use `go env CC` and/or `go env CXX` instead?

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+	MakefileExeTemplate = `# Makefile for python interface for standalone executable package %[1]s.
+# File is generated by gopy. Do not edit.
+# %[2]s
+
+GOCMD=go
+GOBUILD=$(GOCMD) build
+GOIMPORTS=goimports
+PYTHON=%[4]s
+PYTHON_CFG=$(PYTHON)-config
+GCC=gcc
```
sbinet: ditto.

bind/gen.go
```diff
@@ -0,0 +1,666 @@
...
+)
+
+// actually, the weird thing comes from adding symbols to the output, even with dylib!
+// ifeq ($(LIBEXT), ".dylib")
+// 	# python apparently doesn't recognize .dylib on mac, but .so causes extra weird dylib dir..
+// 	- ln -s _%[1]s$(LIBEXT) _%[1]s.so
+// endif
+//
+
+var thePyGen *pyGen
```
sbinet: why do we need this global variable?

bind/gen.go
rcoreilly: just not sure if python is as well-insulated namespace-wise compared to Go -- and lower-case "class" is a keyword so visually, it just seemed like GoClass might be safer?

_examples/hi/test.py
rcoreilly: pybindgen doesn't automatically support varargs.. so it needs the full 2 args per the Go signature.  more generally, retaining the Go signature seems like a good idea, right?

bind/gen.go
rcoreilly: after the PR is accepted, yes, but until then, wouldn't be able to use / test this forked version.. is there any good general soln to this problem?

main_test.go
```diff
@@ -740,46 +563,45 @@ OK
...
+		lang: features[path],
 		want: []byte(`l.SomeString : some string
 l.SomeInt : 1337
 l.SomeFloat : 1337.1337
 l.SomeBool : True
-l.SomeListOfStrings: []string{"some", "list", "of", "strings"}
-l.SomeListOfInts: []int64{6, 2, 9, 1}
-l.SomeListOfFloats: []float64{6.6, 2.2, 9.9, 1.1}
-l.SomeListOfBools: []bool{true, false, true, false}
+l.SomeListOfStrings: go.Slice_string len: 4 handle: 2 [some, list, of, strings]
```
sbinet: we should probably keep some quotes around the elements. either `"` as it used to or `'` (which may be more pythonic.)

main_test.go
```diff
@@ -740,46 +563,45 @@ OK
...
 	testPkg(t, pkg{
 		path: path,
-		lang: []string{"py2", "py2-cffi", "py3-cffi", "pypy2-cffi", "pypy3-cffi"},
-		want: []byte(`sliceptr.IntVector{1, 2, 3}
-sliceptr.IntVector{1, 2, 3, 4}
-sliceptr.StrVector{"1", "2", "3", "4"}
+		lang: features[path],
+		want: []byte(`sliceptr.sliceptr_IntVector len: 3 handle: 1 [1, 2, 3]
+sliceptr.sliceptr_IntVector len: 4 handle: 1 [1, 2, 3, 4]
+sliceptr.sliceptr_StrVector len: 4 handle: 2 [1, 2, 3, 4]
```
sbinet: hum... do we really want this stuttering?

pkgsetup.go
```diff
@@ -0,0 +1,163 @@
...
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/goki/gopy/bind"
```
sbinet: `s/goki/go-python/`

sbinet:
ah, and could you also add the installation of `goimports` to the `.travis.yml` file as well?

sbinet:
FYI, I've added the installation of `goimports` in https://github.com/go-python/gopy/pull/181
(you'll just have to rebase once it's in.)

bind/gen.go
rcoreilly: Any final thoughts about this?  I could go either way..

bind/gen.go
rcoreilly:
```Go
// thePyGen is the current pyGen which is needed in symbols to lookup
// package paths -- not very clean to pass around or add to various
// data structures to make local, but if that ends up being critical
// for some reason, it could be done.
var thePyGen *pyGen
```

bind/gen.go
```diff
@@ -0,0 +1,673 @@
...
+)
+
+// this version uses pybindgen and a generated .go file to do the binding
+
+// for all preambles: 1 = name of package (outname), 2 = cmdstr
+
+// GoHandle is the type to use for the Handle map key, go-side
+// could be a string for more informative but slower handles
+var GoHandle = "int64"
+var CGoHandle = "C.longlong"
```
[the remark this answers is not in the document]
rcoreilly: not sure why you think that?  it is used only in go code, should be fully portable?

bind/gen.go
```diff
@@ -0,0 +1,673 @@
...
+	g.pkg = nil
+	err := os.MkdirAll(g.odir, 0755)
+	if err != nil {
+		return fmt.Errorf("gopy: could not create output directory: %v", err)
+	}
+
+	g.genPackageMap()
+	g.genPre()
+	g.genExtTypesGo()
+	for _, p := range Packages {
```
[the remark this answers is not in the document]
rcoreilly: Packages accumulates over multiple parsing runs for each specific package, and they need to cross-reference, so one would just need to create full outer-loop context for the whole thing, and then create that and destroy it etc.  Not sure there is much practical benefit but is doable in the future if needed.

bind/gen.go
```diff
@@ -0,0 +1,673 @@
...
+func (g *pyGen) genPkgWrapOut() {
+	g.pywrap.Printf("\n\n")
+	g.genPrintOut(g.pkg.pkg.Name()+".py", g.pywrap)
+}
+
+func (g *pyGen) genPkg(p *Package) {
+	g.pkg = p
+	g.pywrap = &printer{buf: new(bytes.Buffer), indentEach: []byte("\t")}
+	g.genPyWrapPreamble()
+	if p == goPackage {
```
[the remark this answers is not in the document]
rcoreilly: for just basic if/ else logic why use a switch?  if there are multiple conditions yes, but I'm not seeing why that would be beneficial here?

bind/gen.go
```diff
@@ -0,0 +1,673 @@
...
+}
+
+func (g *pyGen) genGoPreamble() {
+	pkgimport := ""
+	for pi, _ := range current.imports {
+		pkgimport += fmt.Sprintf("\n\t%q", pi)
+	}
+	pypath, pyonly := filepath.Split(g.vm)
+	pyroot, _ := filepath.Split(filepath.Clean(pypath))
+	libcfg := filepath.Join(filepath.Join(filepath.Join(pyroot, "lib"), "pkgconfig"), pyonly+".pc")
```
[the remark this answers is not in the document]
rcoreilly: filed a ticket.  beyond my expertise.

bind/gen.go
```diff
@@ -0,0 +1,673 @@
...
+			impstr += fmt.Sprintf("import go\n")
+		} else {
+			impstr += fmt.Sprintf("from %s import go\n", g.outname)
+		}
+	}
+	imps := g.pkg.pkg.Imports()
+	for _, im := range imps {
+		ipath := im.Path()
+		if _, has := g.pkgmap[ipath]; has {
+			if g.mode == "gen" || g.mode == "build" {
```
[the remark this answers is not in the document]
rcoreilly: disagree here -- outer logic ecompasses basic if/then inner logic, is more clear this way.

bind/gen.go
```diff
@@ -0,0 +1,673 @@
...
+				impstr += fmt.Sprintf("import %s\n", im.Name())
+			} else {
+				impstr += fmt.Sprintf("from %s import %s\n", g.outname, im.Name())
+			}
+		}
+	}
+
+	if g.mode == "exe" {
```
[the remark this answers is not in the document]
rcoreilly: ditto about basic if / else..

bind/gen.go
```diff
@@ -0,0 +1,673 @@
...
+	names := current.names()
+	for _, n := range names {
+		sym := current.sym(n)
+		if !sym.isType() {
+			continue
+		}
+		if _, has := g.pkgmap[sym.gopkg.Path()]; has {
+			continue
+		}
+		g.genType(sym, true, false) // ext types, no python wrapping
```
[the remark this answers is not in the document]
rcoreilly: IMO worth it for something others will use; not worth it for internal code like this..

bind/gen_func.go
```diff
@@ -0,0 +1,349 @@
...
+	ifchandle, gdoc := isIfaceHandle(gdoc)
+
+	sig := fsym.sig
+	args := sig.Params()
+	res := sig.Results()
+	nargs := 0
+	nres := len(res)
+
+	// note: this is enforced in creation of Func, in newFuncFrom
+	if nres > 2 {
```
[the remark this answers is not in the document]
rcoreilly: yeah but here's where we're unpacking all the vals, so it seems clearer here.  and anyway this is a kind of paranoid check b/c constraint should be enforced earlier.

bind/gen_func.go
```diff
@@ -0,0 +1,349 @@
...
+
+	callArgs := []string{}
+	wrapArgs := []string{}
+	if isMethod {
+		wrapArgs = append(wrapArgs, "self.handle")
+	}
+	for i, arg := range args {
+		na := ""
+		anm := pySafeArg(arg.Name(), i)
+		if ifchandle && arg.sym.goname == "interface{}" {
```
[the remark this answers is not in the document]
rcoreilly: requires shared goArgs in else clause -- would be more complex and less clear to write as switch / fallthrough etc.

gopyh/handle.go
```diff
@@ -0,0 +1,179 @@
...
+package gopyh
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+	"sync"
+)
+
+// type for the handle -- int64 for speed (can switch to string)
```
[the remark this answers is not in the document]
rcoreilly: removed string option entirely at this point -- really only useful for debugging / development.

bind/gen_func.go
```diff
@@ -0,0 +1,349 @@
...
+		g.gofile.Printf("if err != nil {\n")
+		g.gofile.Indent()
+		g.gofile.Printf("C.PyErr_SetString(C.PyExc_RuntimeError, C.CString(err.Error()))\n")
+		if rvIsErr {
+			g.gofile.Printf("return C.CString(err.Error())")
+		}
+		g.gofile.Outdent()
+		g.gofile.Printf("}\n")
+		if rvIsErr {
+			g.gofile.Printf("return C.CString(\"\")")
```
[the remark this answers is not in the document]
rcoreilly: filed ticket for this and other such cases -- I couldn't see any obvious way around leaking it but maybe there are ways..

bind/gen_map.go
```diff
@@ -0,0 +1,353 @@
...
+	qNm := pkgname + "." + slNm
+	typ := slc.GoType().Underlying().(*types.Map)
+	esym := current.symtype(typ.Elem())
+	ksym := current.symtype(typ.Key())
+
+	// key slice type and name
+	keyslt := types.NewSlice(typ.Key())
+	keyslsym := current.symtype(keyslt)
+	if keyslsym == nil {
+		fmt.Printf("nil key slice type!: %s map: %s\n", current.fullTypeString(keyslt), slc.goname)
```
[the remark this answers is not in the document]
rcoreilly: relatively big job as deeply nested in a bunch of functions that don't return errors.. also very rare case I think..

bind/package.go
```diff
@@ -13,41 +13,57 @@ import (
...
-// Package also collects informations about structs and funcs.
+// Package also collects information about specific types (structs, ifaces, etc)
 type Package struct {
 	pkg *types.Package
 	n   int // number of entities to wrap
 	sz  types.Sizes
 	doc *doc.Package
 
-	syms    *symtab
+	syms    *symtab // note: this is now *always* = symbols.current
```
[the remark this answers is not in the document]
rcoreilly: sounds like it is always the case?  wasn't in some prior iteration of the code?

bind/package.go
```diff
@@ -13,41 +13,57 @@ import (
...
 }
 
 // NewPackage creates a new Package, tying types.Package and ast.Package together.
 func NewPackage(pkg *types.Package, doc *doc.Package) (*Package, error) {
 	// protection for parallel tests
 	universeMutex.Lock()
 	defer universeMutex.Unlock()
-	universe.pkg = pkg // FIXME(sbinet)
-
+	fmt.Printf("\n--- Processing package: %v ---\n", pkg.Path())
```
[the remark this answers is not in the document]
rcoreilly: the printf? it is very helpful for localizing any subsequent err messages when processing multi-packages..

bind/stdtypes.go
```diff
@@ -0,0 +1,412 @@
...
+	Packages = append(Packages, goPackage)
+}
+
+// addStdSliceMaps adds std Slice and Map types to universe
+func addStdSliceMaps() {
+	makeGoPackage()
+	gopk := goPackage.pkg
+	sltyps := []string{"int", "int64", "int32", "int16", "int8", "uint", "uint64", "uint32", "uint16", "uint8", "bool", "byte", "rune", "float64", "float32", "string"}
+	for _, tn := range sltyps {
+		universe.addSliceType(gopk, nil, types.NewSlice(universe.sym(tn).gotyp), skType, "Slice_"+tn, "[]"+tn)
```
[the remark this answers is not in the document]
rcoreilly: yep, another day..

bind/symbols.go
```diff
@@ -0,0 +1,1136 @@
...
+	// 	bstr += fmt.Sprintf("\tC.gopy_incref(%s)\n", methvar)
+	// 	bstr += fmt.Sprintf("\t%s = C.PyTuple_New(%d)\n", varnm, sz+1)
+	// 	bstr += fmt.Sprintf("\tC.PyTuple_SetItem(%s, 0, C.PyMethod_Self(%s))\n", varnm, methvar)
+	// 	bstr += fmt.Sprintf("\t_pyargstidx = 1\n")
+	// 	bstr += fmt.Sprintf("\t%[1]s = C.PyMethod_Function(%[1]s)\n", methvar)
+	// 	bstr += fmt.Sprintf("} else {\n")
+	// 	bstr += fmt.Sprintf("\t%s = C.PyTuple_New(%d)\n", varnm, sz)
+	// 	bstr += fmt.Sprintf("}\n")
+	// }
+	bstr := fmt.Sprintf("%s := C.PyTuple_New(%d)\n", varnm, sz)
```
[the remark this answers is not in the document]
rcoreilly: probably most efficient to decouple when / if new backend is implemented -- there are likely various intertwined, extensive such issues and anticipating them all is harder than just dealing with them in concrete case, IMO.

bind/types.go
```diff
@@ -52,3 +53,516 @@ func needWrapType(typ types.Type) bool {
...
+
+func (s *Struct) GoName() string {
+	return s.sym.goname
+}
+
+func (s *Struct) Struct() *types.Struct {
+	return s.sym.GoType().Underlying().(*types.Struct)
+}
+
+// FirstEmbed returns the first embedded type in the class
```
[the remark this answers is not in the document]
rcoreilly: clarified description.  my understanding is that Go convention is to put anonymous embedded types first, which makes sense, and this is optimized for that case. Could go through all such fields but likely unnecessary.

main_test.go
rcoreilly: I believe there are cases where having the package name in the type name is necessary..

bind/gen.go
rcoreilly: added issue..

bind/symbols.go
```diff
@@ -0,0 +1,1136 @@
...
+// typeIdName returns typeGoName with . -> _ -- this should always be used for id
+func typeIdName(t types.Type) string {
+	idn := strings.Replace(typeGoName(t), ".", "_", -1)
+	if _, isary := t.(*types.Array); isary {
+		idn = strings.Replace(idn, "[", "Array_", 1)
+		idn = strings.Replace(idn, "]", "_", 1)
+	}
+	idn = strings.Replace(idn, "[]", "Slice_", -1)
+	idn = strings.Replace(idn, "map[", "Map_", -1)
+	idn = strings.Replace(idn, "[", "Map_", -1)
```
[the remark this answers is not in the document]
rcoreilly: fixed by replacing [ with _ -- confirmed working in GoGi big test case..

main_test.go
```diff
@@ -87,186 +88,59 @@ func TestGofmt(t *testing.T) {
 }
 
 func TestHi(t *testing.T) {
-	t.Parallel()
+	// t.Parallel()
```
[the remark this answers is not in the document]
rcoreilly: couldn't distinguish errors from v2 vs. v3 otherwise.  v2 produces different output fmt vs. v3..

bind/package.go
```diff
@@ -13,41 +13,57 @@ import (
...
+	structs []*Struct
+	ifaces  []*Interface
+	slices  []*Slice
+	maps    []*Map
+	funcs   []*Func
+	// calls   []*Signature // todo: could optimize calls back into python to gen once
+}
+
+// accumulates all the packages processed
+var Packages []*Package
```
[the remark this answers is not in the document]
rcoreilly: per above this is something that could be fixed but would require a fair amount of work to create and maintain and pass a container variable around, and produce no practical benefit that I can see.. there is one method: ResetPackages() that is only needed during running tests, and it is documented as such.

main_test.go
```diff
@@ -293,26 +167,21 @@ hi.Person{Name="foo", Age=42}
...
 --- p.Salary(2): 20
---- p.Salary(24): caught: can't work for 24 hours!
+--- p.Salary(24): caught: <built-in function hi_Person_Salary> returned a result with an error set
 --- Person.__init__
-caught: invalid type for 'Name' attribute | err-type: <type 'exceptions.TypeError'>
-caught: invalid type for 'Age' attribute | err-type: <type 'exceptions.TypeError'>
-caught: Person.__init__ takes at most 2 argument(s) | err-type: <type 'exceptions.TypeError'>
+caught: argument 2 must be str, not int | err-type: <class 'TypeError'>
+caught: an integer is required (got type str) | err-type: <class 'TypeError'>
+*ERROR* no exception raised!
```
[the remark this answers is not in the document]
rcoreilly: yep it apparently doesn't check for *extra* args, just silently ignores them.  it is a bit tricky because not all fields are settable in the constructor so we can't actually check this completely without multiple passes to check etc.  not sure it is worth it.

main_test.go
```diff
@@ -322,73 +191,72 @@ hi.Couple{P1=hi.Person{Name="", Age=0}, P2=hi.Person{Name="", Age=0}}
...
 hi.Couple{P1=hi.Person{Name="mom", Age=50}, P2=hi.Person{Name="bob", Age=51}}
 --- Couple.__init__
-hi.Couple{P1=hi.Person{Name="p1", Age=42}, P2=hi.Person{Name="", Age=0}}
 hi.Couple{P1=hi.Person{Name="p1", Age=42}, P2=hi.Person{Name="p2", Age=52}}
 hi.Couple{P1=hi.Person{Name="p1", Age=42}, P2=hi.Person{Name="p2", Age=52}}
 hi.Couple{P1=hi.Person{Name="p2", Age=52}, P2=hi.Person{Name="p1", Age=42}}
-caught: 'int' object has no attribute 'cgopy' | err-type: <type 'exceptions.AttributeError'>
-caught: 'int' object has no attribute 'cgopy' | err-type: <type 'exceptions.AttributeError'>
-caught: 'int' object has no attribute 'cgopy' | err-type: <type 'exceptions.AttributeError'>
+*ERROR* no exception raised!
```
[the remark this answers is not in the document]
rcoreilly: unfortunately it seems like it is treating those as handles, and given that a handle is an int, the normal pybindgen type checking doesn't catch it.  need to figure out another kind of type check here -- filed an issue.

main_test.go
rcoreilly: i'm not sure how to detect the slice element type in python wrapper properly, that works across versions etc.

sbinet:
I'll go though this tomorrow.
thanks!

nickpoorman:
Any progress on this?

nickpoorman:
I put together a Docker POC for this: https://github.com/nickpoorman/gopy-pr180.

One thing I came across that I covered in the README...

- I'm not sure if this is an issue with `python3.pc` or what, but I had to add `-fPIC --shared` to the generated [Makefile](https://github.com/nickpoorman/gopy-pr180/blob/master/out/Makefile#L35) to get this to build.

sbinet:
ok... I tried to make travis and `go test` happy.
to no avail.

I've applied the Makefile modification from @nickpoorman (thanks!) as well as some gcc arguments cleanup.

the current status of this work is in the [pr-180](https://github.com/go-python/gopy/compare/pr-180?expand=1) branch.

@rcoreilly could you take it from there and make sure travis is green?
(I've dropped `pypy` and `pypy3` for now. and Windows.)

sbinet:
yeah, that's why (and also b/c of Windows) there's this `getPythonConfig` function in `bind/util.go`.
see:
- https://github.com/go-python/gopy/commit/008dc92dd3e1923e6b7be1e7735260f9dfb0f599

perhaps it would need to be refactored into its own package so it could be used from `cmd_build.go` ?

> I'm currently moving so may not be able to do this for a few days.

no worries. (I've just emerged from my email backlog of a 3 weeks holidays...)

bind/package.go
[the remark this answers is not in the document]
justinfx: Logger instead, which allows control of the output?

rcoreilly:
Umm.. it looks like it is now just working!?

https://travis-ci.org/go-python/gopy

sbinet:
that's `master` :)

this is my latest attempt: https://travis-ci.org/go-python/gopy/builds/566516723

rcoreilly:
now that this is all in a branch (pr-180) I don't seem to have push access..  In testing on my linux VM, I was able to replicate the error, and fixed it by setting the `LD_LIBRARY_PATH` to include current dir : .   So adding this after the PATH setting in .travis.yml should hopefully fix it:

```yml
   # ld needs to look in current directory to load the _go.so library
   - export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:.
```

we presumably need to add this to the README.md as well, for anyone running on linux without this set.  somehow mac defaults to using . automatically, so it didn't come up there.

sbinet:
thanks for tackling this (and your persistance).

it's better:
- https://travis-ci.org/go-python/gopy/jobs/570748963

but there are still a couple of failures (some are what seems to be `stdout` buffering issues, others are float/int divisions)

the `pr-180` branch was meant to be cherry-picked from, so you could incorporate my attempts into your branch.

_eg:_
```
$> git checkout your-branch
$> git cherry-pick 9a6a124 a74d9d3 7fa5a17 74d5313 9dc7486 e637039 525b305 13b2ccf 1df5848 008dc92 fb6e893 de9829a 
```

hth,
-s

rcoreilly:
ok.  we will need different "want" outputs for py2 vs. py3 :(

rcoreilly:
I have no idea what this conflict thing is about -- I just edited .travis.yml trivially to resolve, but now it is causing merge problems -- must have been some kind of whitespace?  hopefully you can resolve at final merge?

also, it looks like mac vs. linux has different output -- tests all pass (except v2 vs. v3 diffs) on my mac but not on linux -- I don't suppose you have any experience with that?  should we just target linux given that is what runs on travis?

rcoreilly:
I verified that "go test" passes all tests on linux when only using py3, and filed an issue for adding separate want target for py2.  If this works on travis, then it would be great to just merge this thing and do additional work from there.  It looks like something bad is happening with this conflict (or something else) that is now preventing travis from running:

https://travis-ci.org/go-python/gopy/requests

there is also a possibility that the environment on travis works differently than my "go test" run on my linux vm, in which case we have another problem..

justinfx:
I've been tracking the activity in this pr-180 branch and decided to give the current state a test in my project. My project currently has been using a hand-rolled combination of `github.com/sbinet/go-python` to generated a c-shared library and ctypes to expose the 3 function it needs for passing `*PyObject` in and getting a `*PyObject` back. So to test this pr-180 branch, I split out the main.go into a new packaged without the export go-python binding stuff. I'm left with 3 public functions that I expect to export to python. 
Now I have a couple questions about the results I am getting...

Building using python-2.7, with `go py pkg -name='mypkg' domain.com/group/proj/cmds/gopytest`

It seems a bit fiddly with regards to modules vs $GOPATH. I had to disable GO111MODULE and make sure gopy and my project were in $GOPATH so it could find everything.

The LDFLAGS don't seems to include something like `-Wl,-rpath=$ORIGIN` which means when I try to import the generated pkg, the `_mypkg.so` can find `mypkg_go.so`. So I hand edited the Makefile to add ``-Wl,-rpath='$$ORIGIN'`` and re-run `make build`. Should there be something like this by default or at least a way to inject linker flags?

My project has quite a lot of packages and the cmds/gopytest/main.go has quite a big import chain, so I was pretty suprised to find that I was able to make a successful call through the generated module:

1. Construct a "Request" struct
1. Pass it to the resolve function in my module
1. Get back a "Result" struct type

Now my question is about the fields that are part of the result. 

```go
type Result struct {
    Things []*api.Thing
}
```

I found that my `Result.Things` field did not directly contain the wrapped Thing instances, but rather a slice  `Slice_Ptr_api_Thing([2157, 2158, ...`. I was able to manually map them with something like `[mypkg.Thing(handle=i) for i in result.Things]`. Is that expected? Or ideally would they already be wrapped instead of being a list of long ids?

Also I found that the `api.Thing` type was not exported automatically. I had to type alias it into the main export go file to get it picked up and exported as a class in python. 

Really interested in this branch, and thanks for all the work!

rcoreilly:
@justinfx -- you can specify multiple packages to include, and you might get better results by including the api package directly -- that might cause it to properly interpret the handles as `*api.Thing` instead of having them as bare handles.  If not, could you file a ticket?  It doesn't look like our slice test case tests slices of structs or *struct so we should add that.  I would prefer to do that after the merge instead of adding more to this current pr which seems to be getting a bit unwieldy.

I'm pretty sure most of my current real-world use cases are slices of interface types and those all seem to be working properly without any bare handles being exposed.

Also, adding an additional LDFLAGS would be a good idea. I'm not sure if the current cgo versions of those env vars might already work?  look that up and give it a try if you didn't already.

justinfx:
I ended up creating a small test project and submitting #191 to illustrate the problem with the multiple packages and slice of structs.

> I'm pretty sure most of my current real-world use cases are slices of interface types and those all seem to be working properly without any bare handles being exposed.

Per #191 I tried interfaces but still saw handles :-/

> Also, adding an additional LDFLAGS would be a good idea. I'm not sure if the current cgo versions of those env vars might already work? look that up and give it a try if you didn't already.

Setting `CGO_LD_FLAGS` only influences the linking of "package_go.so" whereas we need to influence the flags for "_package.so". Those compiler flags seem to be derived from the pkg-config python flags. So there would need to be a way to pass more flags into that compiler/linker command that gets built by gopy.

sbinet:
oops, I tried to fixup the merge conflicts using the github web editor but it seems it committed my changes into your `goki/gopy:master` branch.
that's not very polite. sincere apologies.

Issue #191, opened 2019-08-14
Slices of structs are exposed as handles instead of wrapped instances


justinfx:
In response to [my question here](https://github.com/go-python/gopy/pull/180#issuecomment-521049840), I have filed an issue with an example case.

Refs #180 

Test project:
[gopytest.zip](https://github.com/go-python/gopy/files/3503660/gopytest.zip)

Building using python-2.7, with 

```
gopy pkg -name='gopytest' -output=./dist gopytest/pyapi
```

The project contains an "api" package, and then a "pyapi" package meant to export a `Request` capability over the api, to python.

```python
from gopytest import pyapi
req = pyapi.Request()
req.Count=2
ret = req.FetchPeople()
print(ret)
# pyapi.Slice_Ptr_api_Person([6, 7])
```

I find that instead of getting back a list of `Person` instance, I would get a list of handles.
`Person` class is actually not discovered from the import of "api" and exported, so I had to do a type alias in "pyapi" to end up seeing a python `Person` class. Then I was able to manually map the handles to instances by constructing `Person(handle=i)`.

Now if I try building both packages:

```
gopy pkg -name='gopytest' -output=./dist gopytest/pyapi  gopytest/api 
```
while it does generate all the types and two python modules, the api and pyapi modules have circular import dependencies and can't be used properly. "api" module will contain functions that want to call into "pyapi" which has not been imported. And if it were imported manually its a breaking circular import.

(from pr-180)
> I'm pretty sure most of my current real-world use cases are slices of interface types and those all seem to be working properly without any bare handles being exposed.

Based on this suggestion, I tried adding a return type of a slice of interfaces, but it also ended up just creating a python list of handles.

PR #198, opened 2019-09-25
fixes issue #191 (slice elem method auto-wraps the handle, also maps), and fixes imports

rcoreilly:
due to ordering of processing, some incidental types can be created / used in other packages that are not officially included in the go imports -- so we needed to track those and include them in the imports.  took a bit of a hack to write imports at end after having generated entire file, but it works..


rcoreilly:
ps. I don't know why it insists on showing all the old history in this -- the new changes are only the last 6 commits, after merge.  I guess I could just nuke entire repo and refork or something.  instead followed the instructions for merging.

sbinet:
in the spirit of keeping the history as linear as possible, I have created a `pr-198` branch.
see #199.

created with:
```
$> git checkout master
$> git pull
$> git checkout -b pr-198
$> git cherry-pick 6308cc9 25ab0a7 498eba0 c4138aa 85898bc

## 85898bc is the current goki/master
$> git diff 85898bc..1f4c6b6
```

closing this one in favor of the other one.

PR #199, opened 2019-09-30
fixes issue #191 (slice elem method auto-wraps the handle, also maps), and fixes imports

### LINKED ISSUE #191: Slices of structs are exposed as handles instead of wrapped instances

sbinet:
due to ordering of processing, some incidental types can be created / used in other packages that are not officially included in the go imports -- so we needed to track those and include them in the imports. took a bit of a hack to write imports at end after having generated entire file, but it works..

Fixes #191.


sbinet:
thanks @rcoreilly

PR #202, opened 2019-10-20
Fix comments, Examples in README.md

tomoris:
When I tried to execute examples, I got same errors and spent time for the errors. The errors depend on the environment (python env, or LD_LIBRARY_PATH), but I recommend you to revise the document for ease of execution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,11 +176,12 @@ package hi exposes a few Go functions to be wrapped and used from Python.
 
 ### From the command line
 ```sh
-$ gopy build -output=out github.com/go-python/gopy/_examples/hi
+$ gopy build -output=out -vm=`which python3` github.com/go-python/gopy/_examples/hi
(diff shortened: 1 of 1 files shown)


rcoreilly:
The README above this section has the following:

> NOTE: following not yet working in new version:

I never tested this way of running it, and have only used the pkg and exe modes in my usage, except for running the tests through go test, so it might be good to go back and clean this up a bit, and at least remove that comment if it does in fact work.  I'm not really sure what the use-case is for this way of working with things vs. the pkg, exe commands at this point, but presumably that should be documented too.

kushaldas:
> I'm not really sure what the use-case is for this way of working with things vs. the pkg, exe commands at this point, but presumably that should be documented too.

This helps to test any change in the code fast while in the development state, and kind of common habit of many Python programmers.

PR #209, opened 2020-01-23
Added environment variable support

Keithcat1:
Currently, Gopy does not work easily on Windows.
While not automatically fixing that, this allows you to use environment variables to help Gopy find the include and lib dirs.
Also adds -Ofast flag when building Python extension module, in order to help performance a little bit.
Filepath.ToSlashes is also used for include and lib directories as it seems to be required.
The environment variables are:
GOPY_INCLUDE: The include directory where your Python headers are. For example: "C:\python38\include"
GOPY_LIBDIR: The folder where the Python development libraries are. Example: "C:\python38\libs".
GOPY_PYLIB: The name of the Python library to link against. It's what ever you put after the -l command if you manually compile extension modules using GCC. Example: "python37".
These are only used if all three of them are defined.
It is possible to set the environment variables only once, so it doesn't have to be changed every time you open a terminal.
Also should use .pyd extension now.
Right now, I can produce bindings with a package and load it directly from Python, with no need to change anything.

### DIFF
--- a/cmd_build.go
+++ b/cmd_build.go
@@ -6,15 +6,15 @@ package main
 
 import (
 	"fmt"
-	"log"
-	"os"
-	"os/exec"
-	"strings"
-"runtime"
 	"github.com/go-python/gopy/bind"
 	"github.com/gonuts/commander"
 	"github.com/gonuts/flag"
-"path/filepath"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
 )
 
 func gopyMakeCmdBuild() *commander.Command {
@@ -171,11 +171,11 @@ func runBuild(mode bind.BuildMode, odir, outname, cmdstr, vm, mainstr string, sy
 			return err
 		}
 		cccmd := strings.TrimSpace(string(cccmdb))
-		var cflags , ldflags []byte
-		if runtime.GOOS!="windows" {
+		var cflags, ldflags []byte
+		if runtime.GOOS != "windows" {
 			fmt.Printf("%v-config --cflags\n", vm)
 			cmd = exec.Command(vm+"-config", "--cflags") // TODO: need minor version!
-			cflags, err= cmd.CombinedOutput()
+			cflags, err = cmd.CombinedOutput()
 			if err != nil {
 				fmt.Printf("cmd had error: %v  output:\n%v\n", err, string(cflags))
 				return err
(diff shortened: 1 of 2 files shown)


cmd_build.go
```diff
@@ -170,28 +171,49 @@ func runBuild(mode bind.BuildMode, odir, outname, cmdstr, vm, mainstr string, sy
...
-		modlib := "_" + outname + libExt
+		modlib := "_" + outname + extext
 		gccargs := []string{outname + ".c", extraGccArgs, outname + "_go" + libExt, "-o", modlib}
 		gccargs = append(gccargs, strings.Split(strings.TrimSpace(string(cflags)), " ")...)
 		gccargs = append(gccargs, strings.Split(strings.TrimSpace(string(ldflags)), " ")...)
-		gccargs = append(gccargs, "-fPIC", "--shared")
+		gccargs = append(gccargs, "-fPIC", "--shared", "-Ofast")
+		if !symbols { gccargs=append(gccargs, "-s") }
+		include, exists:=os.LookupEnv("GOPY_INCLUDE")
+		if exists==true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `exists` (from `gosimple`)

cmd_build.go
```diff
@@ -170,28 +171,49 @@ func runBuild(mode bind.BuildMode, odir, outname, cmdstr, vm, mainstr string, sy
...
+			gccargs=append(gccargs, "-I"+filepath.ToSlash(include))
+		}
+		lib, exists:=os.LookupEnv("GOPY_LIBDIR")
+		if exists==true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `exists` (from `gosimple`)

cmd_build.go
```diff
@@ -170,28 +171,49 @@ func runBuild(mode bind.BuildMode, odir, outname, cmdstr, vm, mainstr string, sy
...
+			gccargs=append(gccargs, "-L"+filepath.ToSlash(lib))
+		}
+		libname, exists:=os.LookupEnv("GOPY_PYLIB")
+		if exists==true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `exists` (from `gosimple`)

Keithcat1:
Hi,
I fixed that. Thanks.

sbinet:
tests are failing (your edits weren't properly `gofmt`'ed.)
could you please fix that ?
(probably a simple `go fmt .` would suffice)

Keithcat1:
Done.

Keithcat1:
OK, the PR should be sent by the time you see this message.

sbinet:
needs https://github.com/go-python/license/pull/13

PR #213, opened 2020-02-12
fix tests to work with go modules

cosnicolaou:
I noticed that gopy had to be in a GOPATH, pre modules, directory layout for the tests to work. This PR writes a dummy go.mod into the temp dirs created for the tests.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -805,36 +805,17 @@ func testPkg(t *testing.T, table pkg) {
 	}
 }
 
-func copyPackage() (string, error) {
-	var err error
-	pkgCopyDir, err := ioutil.TempDir("", "gopy-")
-	if err != nil {
-		return "", err
-	}
-	target := filepath.Join(pkgCopyDir, "src", "github.com", "go-python", "gopy")
-	if err = os.MkdirAll(target, 0744); err != nil {
-		os.RemoveAll(pkgCopyDir)
-		return "", err
-	}
-	cmd := exec.Command("cp", "-r", ".", target)
-	if err := cmd.Run(); err != nil {
-		os.RemoveAll(pkgCopyDir)
-		return "", err
-	}
-	return pkgCopyDir, nil
-}
+func writeGoMod(t *testing.T, pkgDir, tstDir string) {
+	template := `
+module dummy
 
-func TestMain(m *testing.M) {
-	pkgCopyDir, err := copyPackage()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to install copy of gopy in a temp dir: %v", err)
-		os.Exit(1)
+require github.com/go-python/gopy v0.0.0
+replace github.com/go-python/gopy => %s
+`
+	contents := fmt.Sprintf(template, pkgDir)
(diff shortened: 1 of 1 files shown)


main_test.go
```diff
@@ -805,6 +805,38 @@ func testPkg(t *testing.T, table pkg) {
...
+	pkgCopyDir, err := ioutil.TempDir("", "gopy-")
+	if err != nil {
+		return "", err
+	}
+	target := filepath.Join(pkgCopyDir, "src", "github.com", "go-python", "gopy")
+	if err = os.MkdirAll(target, 0744); err != nil {
+		os.RemoveAll(pkgCopyDir)
+		return "", err
+	}
+	cmd := exec.Command("cp", "-r", ".", target)
```
sbinet: I am surprised this worked on Windows :)

main_test.go
```diff
@@ -805,6 +805,38 @@ func testPkg(t *testing.T, table pkg) {
...
+func TestMain(m *testing.M) {
+	pkgCopyDir, err := copyPackage()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to install copy of gopy in a temp dir: %v", err)
+		os.Exit(1)
+	}
+	testEnvironment = append(testEnvironment, "GOPATH="+pkgCopyDir)
+	fmt.Printf("gopy installed: %v\n", pkgCopyDir)
+	code := m.Run()
+	os.RemoveAll(pkgCopyDir)
```
sbinet: perhaps move it after the `err` check and convert it to a `defer` call?

main_test.go
```diff
@@ -805,6 +805,38 @@ func testPkg(t *testing.T, table pkg) {
...
+		os.RemoveAll(pkgCopyDir)
+		return "", err
+	}
+	return pkgCopyDir, nil
+}
+
+func TestMain(m *testing.M) {
+	pkgCopyDir, err := copyPackage()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to install copy of gopy in a temp dir: %v", err)
```
sbinet:
```suggestion
		fmt.Fprintf(os.Stderr, "failed to install copy of gopy in a temp dir: %+v\n", err)
```

main_test.go
```diff
@@ -805,6 +805,38 @@ func testPkg(t *testing.T, table pkg) {
...
+		os.Exit(1)
+	}
+	testEnvironment = append(testEnvironment, "GOPATH="+pkgCopyDir)
```
sbinet: perhaps add a `FIXME` mentioning we should get rid of this in favor of full Go modules support?
(and/or file an issue about that?)

cosnicolaou: Sorry, I neglected windows! But rather than firing up a vm to test things with the correct incantation of copy I went back to my original approach of writing a fake go.mod. It seems that the documentation for go modules is a little misleading and that a require statement must still accompany a replace, at least for goimports to work. (The docs mention a problem in 1.11 that was fixed in 1.12+ but I think goimports works differently). Anyway, the go.mod approach works now!

Issue #210, opened 2020-02-04
Handles memory is ever growing and causes a memory leak


sb2nov:
https://github.com/go-python/gopy/blob/master/gopyh/handle.go#L35 this map seems to be ever growing. This causes my python process to use more memory indefinitely. 

Can you explain how are handles garbage collected?

rcoreilly:
Yep that is a known problem with the design.  In theory it might be possible to tap into the python garbage collector to delete unused items on the map, but that is probably a pretty tricky thing -- I haven't looked into it though (maybe it would not be too hard?)

In the mean time, trying to organize things so that the python-accessed part remains at a fairly broad outer-loop level would help mitigate this issue.

It might also be useful at some point to revisit some of the slice and map access code to make sure it is doing things as efficiently as possible.

sb2nov:
Thanks for confirming. I was trying to use some go code in a Pyspark map so running it in an outer loop is not possible.

rcoreilly:
this is now fixed and can be closed due to Cosmos Nicolau's fixes, right?

sbinet:
yes. feel free to reopen if that's not the case.

PR #214, opened 2020-02-12
allow for deletion of handles from the map in gopyh.

### LINKED ISSUE #217: bind: restore type-correctness of arguments
before the migration to `pybindgen`, we could detect that arguments to a function or a method where of the expected type:

```go
package pkg
func Foo(s string) { ... }
```

```py
import "pkg"
pkg.Foo(42) ## raises an Exception
```

it seems this feature has been lost and wasn't caught by our testing infrastructure.

cosnicolaou:
This PR addresses #210 and #217. It adds the ability to delete the handles managed by the gopyh package. To do so, it reference counts the use of each handle and deletes handles when that count  goes to zero. In addition, deletion of handles exposed the problem identified by #217 where integer values can be (sometimes incorrectly) treated as handles. The adopted fix is to special case assignment of basic types vs gopy managed objects in struct setters as per the comments in bind/gen_struct.go.

Fixes #217.


rcoreilly:
Overall looks reasonable to me. Basically the Python wrapper objects now have a `__del__` destructor and that de-registers their handle -- seems sensible.  By deleting from the handles map, it allows, but does not compel, the Go GC to delete the object on the Go side, if it is not otherwise being used.  Assuming your tests all check out, then I can't seen any obvious problems with this approach!

I should have thought to look for a Python destructor function but really my detailed experience with Python has mainly been in this project :)

So is there some further Pythonic way we might want to protect access to the handles so they can't be incorrectly used, potentially after deletion?   I might have missed it but I don't think I saw that.  And I can't remember now to what extent any of the generated code might use other object's handles -- that would need to be fixed if so.

cosnicolaou:
I'm not a python expert either:-) Do you know why the API even allows handles to be passed directly to the constructor? The only use I can imagine is to allow multiple python objects to point to the same go object, maybe to share go objects across modules? If I knew the original use case I may be able to come up with a better solution.

rcoreilly:
Maybe if you access a struct-element from a slice, for example, it creates a new python object and passes the handle as the constructor?  I really have been away from this code long enough that I forget how it works!

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/go-python/gopy
 go 1.12
 
 require (
+	github.com/fzipp/gocyclo v0.0.0-20150627053110-6acd4345c835 // indirect
```
sbinet: not sure we actually need this as an explicit dep.
remove?

go.sum
```diff
@@ -1,3 +1,5 @@
+github.com/fzipp/gocyclo v0.0.0-20150627053110-6acd4345c835 h1:roDmqJ4Qes7hrDOsWsMCce0vQHz3xiMPjJ9m4c2eeNs=
```
sbinet: not sure we actually need this as an explicit dep.
remove?

_examples/hi/test.py
```diff
@@ -168,100 +168,107 @@
...
 
+# It's dangerous to refer to go handles as constants now that they
+# are gc'ed.
+anon1 = hi.Anon()
+h1 = anon1.handle
+h2 = anon.handle
+
 try:
-    hi.Couple(1)
+    hi.Couple(h1)
```
sbinet: I don't understand this change.
originally, this was testing type-correctness.

_examples/hi/test.py
```diff
@@ -168,100 +168,107 @@
...
     print("*ERROR* no exception raised!")
 except Exception as err:
-    print("caught:", err, "| err-type:",type(err))
+    print("caught:", err, "| err-type:", type(err))
     pass
 
 try:
-    hi.Couple(1,2)
+    hi.Couple(h1, h2)
```
sbinet: I don't understand this change.
originally, this was testing type-correctness.

_examples/hi/test.py
```diff
@@ -168,100 +168,107 @@
...
     print("*ERROR* no exception raised!")
 except Exception as err:
-    print("caught:", err, "| err-type:",type(err))
+    print("caught:", err, "| err-type:", type(err))
     pass
 
+
 try:
-    hi.Couple(P2=1)
+    hi.Couple(P2=h1)
```
sbinet: I don't understand this change.
originally, this was testing type-correctness.

_examples/hi/test.py
```diff
@@ -15,20 +15,20 @@
 print("--- hi.Universe:", hi.Universe)
 print("--- hi.Version:", hi.Version)
 
-print("--- hi.Debug():",hi.Debug())
+print("--- hi.Debug():", hi.Debug())
```
sbinet: could you factor out all these cosmetic changes (that I generally agree with) into a dedicated commit ?
(it will be easier for diffs.)

cosnicolaou:
I see the uses of handle= in the generated code that rcoreilly alludes to. I'll add refcounting to allow for that.

_examples/hi/test.py
sbinet: well... this must be some uncaught regression.
(I haven't bisected but I suspect that's since the migration to `pybindgen`)
it used to be that invalid arguments would be caught and an exception be raised.
(I agree, printing `*ERROR* isn't a very robust mechanism...)

sbinet: so, to circle back on this.
I'd rather have the original test being commented out (with the `hi.Couple(1)` left as is) instead of changing the purpose of the test.

I have filed https://github.com/go-python/gopy/issues/217 so we don't forget about this.

cosnicolaou: yep, that's my plan too. I looked back through the git history and can see what happened, there's a TODO in the code. I'll hopefully get this cleaned up today or tomorrow.

cosnicolaou: ok, please take another look, especially at the comments in gen_struct. I think those changes make sense, but the underlying problem is that the confusion between int values that are handles and those that are not and I don't see any clean/easy way of resolving that beyond what I've done here. Let me know if you have any other ideas.

bind/gen.go
```diff
@@ -108,6 +108,25 @@ func GoPyInit() {
 type GoHandle %[4]s
 type CGoHandle %[5]s
 
+// DecRef decrements the reference count for the specified handle
+// and deletes it it goes to zero.
```
sbinet:
```suggestion
// and deletes it if it goes to zero.
```

bind/gen_struct.go
```diff
@@ -53,23 +53,30 @@ in which case a new Go object is constructed first
...
 	for i := 0; i < numFields; i++ {
 		f := s.Struct().Field(i)
 		if _, err := isPyCompatField(f); err != nil {
 			continue
 		}
-		// TODO: this will accept int args for any handles / object fields
-		// need some kind of additional type-checking logic to prevent that in way
-		// that also allows valid handles to be used..
+		// NOTE: this will accept int args for any handles / object fields so
+		// some kind of additional type-checking logic to prevent that in way
```
sbinet:
```suggestion
		// some kind of additional type-checking logic to prevent that in a way
```

bind/gen_struct.go
```diff
@@ -53,23 +53,30 @@ in which case a new Go object is constructed first
...
+		// that also allows valid handles to be used is required. This is
```
sbinet:
```suggestion
		// that also allows valid handles to be used as required. This is
```

bind/gen_struct.go
```diff
@@ -53,23 +53,30 @@ in which case a new Go object is constructed first
...
+		// achieved in the per-field setters (see below) which check to ensure
```
sbinet:
```suggestion
		// achieved in the per-field setters (see below) with checks to ensure
```

bind/gen_struct.go
```diff
@@ -53,23 +53,30 @@ in which case a new Go object is constructed first
...
+		// that a struct field that is an gopy managed object is only
```
sbinet:
```suggestion
		// that a struct field that is a gopy managed object is only
```

bind/gen_struct.go
```diff
@@ -53,23 +53,30 @@ in which case a new Go object is constructed first
...
+		// assigned gopy managed objects. Fields of basic types (e.g int, srring)
```
sbinet:
```suggestion
		// assigned gopy managed objects. Fields of basic types (e.g int, string)
```

gopyh/handle.go
```diff
@@ -102,14 +114,68 @@ func Register(typnm string, ifc interface{}) CGoHandle {
...
+		if cnt < 0 {
+			panic(fmt.Sprintf("gopy DecRef ref count %v for handle: %v, ifc %v", cnt, ghc, handles[ghc]))
+		}
+		if trace {
+			fmt.Printf("gopy DecRef: %d: %d\n", handle, cnt)
+		}
+	}
+}
+
+//  IncRef increments the referene count for the specified handle.
```
sbinet:
```suggestion
//  IncRef increments the reference count for the specified handle.
```

gopyh/handle.go
```diff
@@ -102,14 +114,68 @@ func Register(typnm string, ifc interface{}) CGoHandle {
...
+	defer mu.Unlock()
+	if handles == nil {
+		return
+	}
+	ghc := GoHandle(handle)
+	if _, exists := handles[ghc]; !exists {
+		return
+	}
+	counts[ghc]--
+	if cnt := counts[ghc]; cnt == 0 {
```
sbinet: I'd personally go with a `switch`:
```go
switch cnt := counts[ghs]; {
case cnt == 0:
    ...
case cnt < 0:
    panic(...)
default:
    if trace { ... }
}
```

main_test.go
```diff
@@ -831,7 +864,7 @@ func testPkgBackend(t *testing.T, pyvm string, table pkg) {
 	if err != nil {
 		t.Fatalf("[%s:%s]: could not create workdir: %v\n", pyvm, table.path, err)
 	}
-	defer os.RemoveAll(workdir)
+	//defer os.RemoveAll(workdir)
```
sbinet:
```suggestion
	defer os.RemoveAll(workdir)
```

gopyh/handle.go
cosnicolaou: agreed!

PR #230, opened 2020-09-09
for slices with non-pointer struct element types, need to return pointer to element

rcoreilly:
Pointer is needed so that elements can be modified -- no other way to do this python-side, and generated code assumes pointer receivers.

This has been fully tested on my extensive codebase..

### DIFF
--- a/bind/gen.go
+++ b/bind/gen.go
@@ -250,6 +250,7 @@ os.chdir(cwd)
 # %[2]s
 
 import _%[1]s, collections
+from enum import Enum
 
 # to use this code in your end-user python file, import it as follows:
 # from %[1]s import %[3]s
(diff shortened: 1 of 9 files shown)


rcoreilly:
I will have at least one more additional change to add to this set -- getting ready to more fully support python interface to our neural network simulator, so fixing various issues here.

rcoreilly:
Ok, I think that's it for now -- have successfully converted a Go main program to run essentially identically in Python.

I also have a Go-to-Python converter program mostly working: https://github.com/goki/gotopy  -- it would be appropriate to host it here on go-python if you would be interested.  It might make sense to just give me write permission if so, so I can more efficiently contribute to managing both packages.

sbinet:
done: you're an owner of `go-python` :)

rcoreilly:
Great!  So do you want to review this PR or should I just merge it?

sbinet:
go ahead.
I must admit I have less "free" time on my hands for `gopy` these days...

PR #242, opened 2021-03-13
Update cmd_build.go

hhuangwx:
Add an option for the linker to find the shared object in the current path. This helps to avoid the setting of LD_LIBRARY_PATH


justinfx:
@hhuangwx  are you still working on a PR for this feature? I also want the functionality and will want to work on a PR if you don't have one ready to go?

PR #243, opened 2021-03-14
Create cmd_build.go

hhuangwx:
Add -Wl,-rpath=. to solve the error that shared objects are not found

### DIFF
--- a/cmd_build.go
+++ b/cmd_build.go
@@ -230,7 +230,7 @@ func runBuild(mode bind.BuildMode, odir, outname, cmdstr, vm, mainstr string, sy
 		modlib := "_" + outname + extext
 		gccargs := []string{outname + ".c", extraGccArgs, outname + "_go" + libExt, "-o", modlib}
 		gccargs = append(gccargs, strings.Fields(pycfg.AllFlags())...)
-		gccargs = append(gccargs, "-fPIC", "--shared", "-Ofast")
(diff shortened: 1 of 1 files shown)


justinfx:
What a coincidence! I was actually about to tackle this feature next after my recent PRs!

cmd_build.go
```diff
@@ -230,7 +230,7 @@ func runBuild(mode bind.BuildMode, odir, outname, cmdstr, vm, mainstr string, sy
 		modlib := "_" + outname + extext
 		gccargs := []string{outname + ".c", extraGccArgs, outname + "_go" + libExt, "-o", modlib}
 		gccargs = append(gccargs, strings.Fields(pycfg.AllFlags())...)
-		gccargs = append(gccargs, "-fPIC", "--shared", "-Ofast")
+		gccargs = append(gccargs, "-Wl,-rpath=.", "-fPIC", "--shared", "-Ofast")
```
justinfx, line 233: I think it would be more correct to use `$ORIGIN` instead of `.`
That way it is specifically tied to the same directory as the python extension.

justinfx:
I've submitted a replacement for this merge, via #250

cmd_build.go
hhuangwx: Thank you for thinking it more comprehensively.

hhuangwx:
If everything goes well, this PR can be closed.

sbinet:
superseded by #251

PR #241, opened 2021-03-13
main,bind,_examples: fix tests for go 1.16

justinfx:
This CL fixes a number of tests that are failing under go 1.16.
* module dependencies behave a bit differently
* the `//gopy:` directives now needs a leading space
* generated version file needs go fmt
* there was a mis-named symbol

PR #244, opened 2021-03-14
bind: ensure enum.Enum is imported in generated code as needed

justinfx:
This fix removes the static import of `enum.Enum` that was added only to the exe preamble, and replaces it with a conditional import added to all generated packages, only if there is at least one enum to generate.
The conditional import improves python2 legacy support where `Enum` is an external dependency that must be present.

Depends on ~~#241~~ being merged for test refactors.

### DIFF
--- a/bind/gen.go
+++ b/bind/gen.go
@@ -250,7 +250,6 @@ os.chdir(cwd)
 # %[2]s
 
 import _%[1]s, collections
-from enum import Enum
 
 # to use this code in your end-user python file, import it as follows:
 # from %[1]s import %[3]s
@@ -776,6 +775,10 @@ func (g *pyGen) genAll() {
 	}
 
 	g.pywrap.Printf("\n\n#---- Enums from Go (collections of consts with same type) ---\n")
+	// conditionally add Enum support because it is an external dependency in py2
+	if len(g.pkg.enums) > 0 {
+		g.pywrap.Printf("from enum import Enum\n\n")
+	}
 	for _, e := range g.pkg.enums {
 		g.genEnum(e)
 	}


justinfx:
Rebased to master to fix merge conflict

.gitignore
```diff
@@ -0,0 +1,3 @@
+# IDE files
```
sbinet, line 1: why are these modifications in the CL still?
it seems to me they were already merged in as part of #241 ?

.travis.yml
```diff
@@ -1,9 +1,8 @@
 language: go
 
 go:
-  - 1.13.x
-  - 1.12.x
-  - 1.11.x
+  - 1.16.x
```
sbinet, line 4: ditto

.travis.yml
```diff
@@ -31,8 +30,8 @@ before_install:
    # pypy3 isn't packaged in ubuntu yet.
    - TEMPDIR=$(mktemp -d)
    - export PYPYVERSION=v7.1.1
-   - curl -L https://bitbucket.org/pypy/pypy/downloads/pypy2.7-${PYPYVERSION}-linux64.tar.bz2 --output $TEMPDIR/pypy2.tar.bz2
-   - curl -L https://bitbucket.org/pypy/pypy/downloads/pypy3.6-${PYPYVERSION}-linux64.tar.bz2 --output $TEMPDIR/pypy3.tar.bz2
+   - curl -L https://downloads.python.org/pypy/pypy2.7-${PYPYVERSION}-linux64.tar.bz2 --output $TEMPDIR/pypy2.tar.bz2
+   - curl -L https://downloads.python.org/pypy/pypy3.6-${PYPYVERSION}-linux64.tar.bz2 --output $TEMPDIR/pypy3.tar.bz2
```
sbinet, line 34: ditto.

main_test.go
```diff
@@ -912,7 +917,7 @@ func testPkgBackend(t *testing.T, pyvm string, table pkg) {
 	if err != nil {
 		t.Fatalf("[%s:%s]: could not create workdir: %v\n", pyvm, table.path, err)
 	}
-	defer os.RemoveAll(workdir)
+	//defer os.RemoveAll(workdir)
```
sbinet, line 920: could we retain the defer in place?
(I assume you removed it for debugging)

perhaps modify it to only apply the `os.RemoveAll` if everything went fine?

sbinet (commented):
it seems to me the rebase didn't quite work because I still see some commits from your earlier work of #241

sbinet (approved):

Issue #239, opened 2021-01-03
suggestion for package  install to `fix the import bug`


zju3351689:
suggestion for package  install to `fix the import bug`

with  ‘go build’  to produce hi.py go.py in the  `outputhi` path.

code need changed: 
in ‘go.py’
    `import _hi`     => from . import _hi
    `import go `    => from . import go
in ‘hi.py’
     `import _hi`     => from . import _hi
     `import hi`       => from . import hi
    `import go`      => from .  import go

then the PYTHONPATH could be use the `$PATH/lib/python3.7/site-packages/` not the  `$PATH/lib/python3.7/site-packages/outputhi`

justinfx:
This would be great to fix. I started testing a project with gopy today and hit thag import issue straight away. Had to fix it with a hack to temporarily modify the `sys.path` before and after importing the package

rcoreilly:
Happy to incorporate a PR on this but I don't think I have the bandwidth or domain knowledge to evaluate directly -- paths and python are out of my experties..

justinfx:
This has actually also been a long standing issue across many tickets in the protocol buffers python/grpc code gen: https://github.com/protocolbuffers/protobuf/issues/1491
The same issue occurs in the generated python code regarding import statements and python3 using absolute imports instead of python2 relative imports.

I would be fine to look at working up a patch. But similar to the protobuf issue, I would like to hear an opinion as to whether it should be wholesale changed to `from . import go.py` or if it should be an option like `--import-package=<NAME>`:

```
# absolute
--import-package='mypackage.foo.py'
# relative
--import-package='.'
```

And if the flag were defined, it would end up generating `from <package> import go.py`.
Seeing as python2 is long deprecated and absolute imports are the default in python3, I would opt for making the flag default to "." but allowing some control over the value if needed, in cases where the files might be separated.

PR #245, opened 2021-03-14
main,bind,_examples: Use absolute package imports and make configurable

justinfx:
This CL fixes an issue with the way python import statements are generated using a relative current directory approach that is no longer compatible with python3. The default behavior has been updated to default to a dot-relative absolute package import for "go" and compiled extension, and a configuration option has been exposed from the CLI flags down to the bind package.
A refactor of configuration options was added to avoid further expanding the number of arguments being passed through the system.

Includes #244 enum import refactor

Fixes go-python/gopy#239


justinfx:
Rebased to master to fix merge conflicts

justinfx:
This MR is a bit dirty since the fixes from #241 were merge squashed to master. Will fixup now...

bind/bind.go
```diff
@@ -12,6 +12,22 @@ import (
 	"os"
 )
 
+// BindCfg is a configuration used during binding generation
+type BindCfg struct {
+	// output directory for bindings
+	OutputDir string
+	// "name of output package (otherwise name of first package is used)
```
sbinet, line 19:
```suggestion
	// name of output package (otherwise name of first package is used)
```

bind/bind.go
```diff
@@ -12,6 +12,22 @@ import (
...
+	Name string
+	// code string to run in the go main() function in the cgo library
+	Main string
+	// the full command args as a string, without path to exe
+	Cmd string
+	// path to python interpreter
+	Vm string
```
sbinet, line 26:
```suggestion
	VM string
```
(to follow Go's take on acronyms like `id/ID` vs `Id`...)

sbinet (approved):
LGTM.

thanks.

(I'll wait for #244 to be merged before merging that one as well.)

bind/bind.go
justinfx: Done: 93b1b94

sbinet:
could you rebase off the latest master?
I've migrated to GHA (and dropped Travis-CI), this should speed up the CI turn over time by a lot.

justinfx:
I've rebased but it looks like the windows tests are still appveyor? Is moving those to GHA still wip?

sbinet:
I've indeed just migrated the Linux tests to GHA (as they were the ones in need of some oomph)

rcoreilly:
probably not intentional -- can you submit a PR to fix?

PR #250, opened 2021-03-17
main: Use RPATH=$ORIGIN linker flag with python extension

justinfx:
This commit adds a linker flags to the python extension to set the RPATH=$ORIGIN. Doing so removes the need to set LD_LIBRARY_PATH in both the runtime environment, and in the tests environment, in order for the python extension to find the neighboring Go shared library.

Replaces #243

PR #251, opened 2021-03-18
bind: reduce shared library count from 2 to 1 and use ABI version tagged extension

### LINKED ISSUE #249: Output python extension should use py3+ ABI version tagged naming conventions (PEP 3149)
Given the following build command:

```
gopy build -name=goplugin ...
```

A python extension will be compiled with the name `_goplugin.so`. This is normal for python2, and may be fine for python3 if you never need to build the extension for any other minor versions of python3. But in order to support multiple build variants in a single directory, python3+ builds should make use of ABI version tagged naming: https://www.python.org/dev/peps/pep-3149/

```
# python2.7
python -c 'from distutils.sysconfig import get_config_var; print(get_config_var("EXT_SUFFIX"))'
#

# python3.7
python -c 'from distutils.sysconfig import get_config_var; print(get_config_var("EXT_SUFFIX"))'
# .cpython-37m-x86_64-linux-gnu.so

# or python3.7
python3-config --extension-suffix
# .cpython-37m-x86_64-linux-gnu.so
```

The build process that compiles the python extension should be updated to produce something like this:
```
_goplugin.cpython-37m-x86_64-linux-gnu.so
```

And then it should be safe to build multiple times into the same target output directory, given that the Go shared library and the generated python source files should be the same either way.

justinfx:
This merge replaces the python extension linked against the go shared library with a single python extension module. The extension module is also named using the ABI version tagging reported by the python configuration to avoid clashes with other python minor version builds.
Removes need/reference for LD_LIBRARY_PATH modification.

Fixes #249 
If merged, replaces the need for #250 since we no longer need RPATH linking between 2 shared libraries.

Open question: I have not added anything in the `exe` build case and I am not really sure if this should be added.


sbinet (approved):
LGTM.

(I'll merge this one in lieu of #250.)

sbinet:
as for the `exe` case, I think @rcoreilly is the one with the best judgement call there :)

rcoreilly:
The `exe` directly links in all the go-specific compiled code that would otherwise be in the shared library, so this should not be an issue for that case.  As is the case with go programs in general, the go code is statically linked into the executable, so no shared lib of that code to deal with.  It still does have to find the Python shared lib, but I guess this doesn't deal with that issue.

PR #252, opened 2021-03-20
main,bind: Extended gopy:name control, auto PEP8 naming, and property docstring support

justinfx:
This merge improves the support for controlling the generates names of functions, methods, and properties in the python code. The "gopy:name" tag support has been extended to struct fields and updated to also support single line comments:

```go
type MyStruct struct {
    // gopy_name: my_field
    // This is my field.
    MyField int

    // This is my second field
   // gopy_name: my_field2
   MyField2 int
}
```

A new optional feature flat `-rename` has been added to apply automatic PEP8 snake_case name transformations to functions, methods, properties, and var getter/setters.  This makes it more convenient than having to annotate every symbol in Go source with a `gopy:name` tag if we only want CamelCase converted to snake_case to be more "pythonic".

And lastly, as part of parsing the comments for struct fields, the docstring is actually also applied to the python property getter in the generated code.

Objectives:

1. Is the `rename` flag and auto naming feature useful and complete enough?
2. Is the approach to parsing the struct field comments correct? I tried to review the existing way that method comments are propagated but it seemed fields were a bit different.


_examples/rename/rename.go
```diff
@@ -5,15 +5,38 @@
 // package rename tests changing the names of methods and functions
 package rename
 
-// gopy:name say_hi
+// gopy:name say_hi_fn
+// Comment follows the tag and function shoudl be renamed
```
sbinet, line 9:
```suggestion
// Comment follows the tag and function should be renamed
```

_examples/rename/rename.go
```diff
@@ -5,15 +5,38 @@
...
+// to say_hi_fn
 func SayHi() string {
 	return "hi"
 }
 
+// I should be renamed tp auto_renamed_func, when generated
```
sbinet, line 15:
```suggestion
// I should be renamed to auto_renamed_func, when generated
```

sbinet (commented):
for struct fields, there's perhaps an argument about using struct-tags.
_e.g.:_
```go
type T struct {
    MyField int32 `gopy:"my_field"`
    Hidden  int32 `gopy:"_hidden"`
}
```

the advantage of using a `//gopy:name` directive might be on the godoc side (I suspect that kind of comment may not be displayed so these kind of internal details aren't leaked to the world, contrary to the struct-tag approach).

@rcoreilly ? any input?

justinfx:
@sbinet in hindsight I should have used struct tags. It makes more sense and probably would have saved me the trouble of trying to locate the comments for the field. I was only thinking of making them the same as the other types, but a struct tag is better suited for settings like this that would be removed from the doc string.
Let me work up another commit to try and change this.

rcoreilly:
Yeah I agree that on balance the field tag makes more sense.  All kinds of impl stuff is leaked through those so I don't think this is any different, and is idiomatic in my experience.

justinfx:
I've now added struct tag parsing of `gopy:"new_name"`. Please let me know if you want me to remove the doc comment parsing logic entirely, or leave it in? I left it just for consistency with the other types, but the tag wins.

rcoreilly:
the Go style is to have only one way to do things where possible, so probably remove it..

justinfx:
I've removed the property getter/setter doc parsing for a name override. But I have left the struct field `getDoc` updates because they are needed to pull the comments to then set the property docstring in python

bind/utils.go
```diff
@@ -192,19 +193,82 @@ func getGoVersion(version string) (int64, int64, error) {
...
+// a new python name. If the tag is not defined then the original
+// name is returned.
+// If the tag name is specified but is an invalid python identifier,
+// then an error is returned.
+func extractPythonNameFieldTag(gname, tag string) (string, error) {
+	const tagKey = "gopy"
+	if tag == "" {
+		return gname, nil
+	}
+	tags, err := structtag.Parse(tag)
```
sbinet, line 239: bringing this whole new dependency seems a bit overkill for something that could -I think- be achieved with just:
```go
v, ok := reflect.StructTag(tag).Lookup(tagKey)
```
couldn't it?

justinfx: Oops. I should have paid closer attention to what was already available in the `reflect` package. Fixed!

sbinet: no worries :)

sbinet (approved):
LGTM.

(I'll wait a bit for @rcoreilly before merging this one in)

rcoreilly:
Yep looks good to me too -- thanks for the contributions!

Issue #254, opened 2021-04-04
"returned a result with an error set" instead of Exception for functions that return a value type


justinfx:
I've encountered a situation where a certain wrapped function fails in the Go runtime, but the exception is swallowed and instead a generic Python exception is raised:

```
SystemError: <built-in function mymodule.MyFunction> returned a result with an error set
``` 

We can see an exiting example in the `gopy/_examples/pyerrors` test:

```go
func Div(i, j int) (int, error) {
	if j == 0 {
		return 0, errors.New("Divide by zero.")
	}
	return i / j, nil
}
```

```python
def div(a, b):
    try:
        r = pyerrors.Div(a, b)
        print("pyerrors.Div(%d, %d) = %d"% (a, b, r))
    except Exception as e:
        print(e)

div(5,0)
# <built-in function pyerrors_Div> returned a result with an error set
```

This makes it pretty difficult to debug since one would have to track down the wrapped code and figure out which return statement is actually being hit, then add some logging to see the real error.

I've been digging through pygo and PyBindGen for a few hours to better understand this problem. It seems like @rcoreilly was already aware of this problem when the test was written, and coded around it to have this output actually be a passing test, with a TODO to fix it later?  But it seems like there is a bit of a disconnect between the location where gopy generates the C code with python `C.PyErr_SetString` error handling (`bind/gen_func.go`) vs where PyBindGen accepts C functions being added. I don't understand where there is any existing PyBindGen API to indicate what return value should be considered an error and when to check for one. 

Problem: If a Python error has been set, then a `PyObject*` should not be returned. gopy sets the error, but PyBindGen has no current logic in the generated C code that checks `PyErr_Occurred()` other than some special int pointer case. 

I have no experience with PyBindGen, but a decent amount with Cython. In cython there are a few different ways where you can indicate that a wrapped function can raise an exception. You can tell it an exact value that indicate an exception (ie -1 always being an exception) or you can tell it to always check `PyErr_Occurred` at the extra expense of that check for each call. I don't know how to do that for PyBindGen but ultimately I think we need to figure out how to set that. 

Here is an example of what PyBindGen generates:

```python
# build.py
mod.add_function('mypackage_DoThing', retval('int64_t'), [param('int64_t', '_handle')])
```

```C
PyObject *
_wrap__mypackage_DoThing(PyObject * PYBINDGEN_UNUSED(dummy), PyObject *args, PyObject *kwargs)
{
    PyObject *py_retval;
    int64_t retval;
[...]

PR #255, opened 2021-04-04
bind: Check for exceptions and propagate errors back to python caller

justinfx:
This is  WIP attempt to fix the issue of errors not being propagated back to the python caller as a proper exception. The problem occurs because the wrapped function will check for an error from the Go calls and use `PyErr_SetString` but still return a value. Then the code generated by PyBindGen will only get the retval and return it. This results in:

```
SystemError: <built-in function mymodule.MyFunction> returned a result with an error set
```

The fixes in this PR add a custom `Function` wrapper to the PyBindGen code that adds the function wrappers. In this `Function` class we update the `generate_call()` to do an error check after the wrapped function call. The result of this fix is that the tests that had previously been set up to expect the generic `SystemError` have now been updated to expect the real Go error as an exception.

Known issues:

1. Each function wrapper now checks `PyErr_Occurred()` before returning. I've added the ability for a custom `retval` check first if performance is a concern, but I don't yet know where that expression can be added. As an example, the `CgoHandle` long value uses a `-1` when its not valid. So the failure expression for the function *could* be `retval == -1` which would check that first before calling `PyErr_Occurred()`. This may not be worth the effort.
2. ~~We now could have a potential memory leak during an exception. As an example, a wrapped function that returns a string will generate code that returns `C.CString("")` on error. Before this PR the object is always returned as a python object which then takes ownership of the `char*`. If we now do an error check and return `NULL` then there is not yet any cleanup code that will free that `char*`. I am not yet sure how to add the cleanup code. Ideally on error we would not allocate an empty `char*` and just return `NULL`. It may be easier to fix that part of the code gen then to try and generate C wrapper cleanup code for the `retval`.~~

Refs #254


justinfx:
I thought about the issue of leaked strings during the exception handling and realized that the entire patch-leak generator tool could be removed, and replaced with some more PyBindGen custom Function logic. I've added another commit that cleans up strings both during an exception, and at the end where the patch-leak tool previous injected a `free`. Now an entire build step has been removed.

bind/gen.go
```diff
@@ -203,9 +203,44 @@ func GoPyMainRun() {
...
+def add_checked_function(mod, name, retval, params, failure_expression='', *a, **kw):
+    fn = CheckedFunction(name, retval, params, *a, **kw)
+    fn.set_failure_expression(failure_expression)
+    mod._add_function_obj(fn)
+    return fn
+
+def add_checked_string_function(mod, name, retval, params, failure_expression='', *a, **kw):
+    fn = CheckedFunction(name, retval, params, *a, **kw)
+    cleanup = 'free(retval);'
+    fn.set_failure_cleanup(cleanup)
```
justinfx, line 238: This is doing a `free(retval)` when an error has been detected. But is it safe to assume we will always keep allocating an empty string even on exception, in the Go code? Should I change this to:
```C
if (retval != NULL) free(retval);
```

justinfx:
@rcoreilly might need you for this review since its highly related to PyBindGen

bind/gen.go
justinfx: commit 57a66d6

sbinet (approved):
LGTM.

I'll leave a couple of days for @rcoreilly to chime in.

rcoreilly:
This all seems good -- I didn't really look into it in great depth, but I checked out the PR and built it and looked at the changes and it makes sense locally.

Unfortunately I'm not able to test it fully because `gonum/plot` decided to use the package name `font` which conflicts when you also happen to be using the standard `image/font` -- unlike most go code, gopy puts everything into one ginormous .go file, so these package name conflicts are a problem.  Presumably it would be possible to fix with a package alias but I don't have much time to work on that right now..

Meanwhile, something is not working with the previous import changes.  I'm getting this error:

```
oreilly@shadow:~/goki/gi/examples/widgets/ > ./widgets.py 
Traceback (most recent call last):
  File "/Users/oreilly/go/src/github.com/goki/gi/examples/widgets/./widgets.py", line 16, in <module>
    from gi import go, gi, gist, giv, units, ki, mat32, gimain
  File "/usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/gi/go.py", line 12, in <module>
    from . import _gi
ImportError: cannot import name '_gi' from 'gi' (/usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/gi/__init__.py)
```

I'll look into this later, but anyway you can probably just merge this PR for now.

rcoreilly:
ps. @justinfx if you happen to have a quick fix for how the new relative imports are supposed to work when you install library, that would be great!

justinfx:
@rcoreilly do you have some more details about the specific problem you hit with imports? I wasn't clear if it was a conflict in the installation path or the imports. But the change I made was to go *from* relative imports like `import mylib` (which expect to check the current directory in py2) to absolute imports like `from . import mylib` (which expect to be part of a package that is on your PYTHONPATH). 
Have you tried setting the `gopy build -package-prefix="."` value to something suitable for your package installation location? The default "." assumes the common case that your generated python source file is in the same directory as the compiled extension, within a package. Also maybe the `-name` flag might help to avoid clashes?

justinfx:
Thanks for the review!

PR #260, opened 2021-07-21
Properly set up CGO_CFLAGS and CGO_LDFLAGS on Windows

Keithcat1:
Set up the CGO CFLAGS and CGO LDFLAGS environment variables before building pkg_go.dll.
Also tell Gopy to link directly to python*.dll on Windows. Don't use import libraries because it's hard to reliably figure out where they are. You should be able to just have it work with any Python installation.


rcoreilly:
Could you give the latest version a try and see if that works for you?  If not, please make a new PR on current code.  My strategy was to use the include dir which seems to be reliably found on windows, as a basis for libdir.  Also to recreate the python library name based on major and minor version numbers.

PR #275, opened 2022-04-15
upgrade go.mod make osx(arm) generate file success

Theodoree:


Theodoree:
command: 
```
gopy pkg -version="0.0.1" -name="pkg_py" -output=out -vm=$(which python3)  $lib
```
then will generate slice_xxx function. but not generate $lib function

Theodoree:
probably because the minimum supported version for arm is 1.16

rcoreilly:
I'm not sure why the appveyor is now failing -- @sbinet any ideas about that?  Looks like a mismatch between tools and go version?

AlmogBaku:
hi, any update on this?
go 1.18 is not usable without upgrading the x/tools version

AlmogBaku:
hi, any update on this?
go 1.18 is not usable without upgrading the x/tools version

PR #282, opened 2022-06-15
[Draft] Support for multiple return values from go function to python

nishantsharma:
Support for multiple return values from go function to python

### DIFF
--- a/bind/gen_func.go
+++ b/bind/gen_func.go
@@ -194,7 +194,7 @@ func (g *pyGen) genFuncSig(sym *symbol, fsym *Func) bool {
 			gorets[i] = sret.cgoname
 			pyrets[i] = "'" + sret.cpyname + "'"
 			if sret.cpyname == "PyObject*" {
-				ownership = "caller_owns_return=True"
+				ownership = ", caller_owns_return=True"
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


nishantsharma:
Some memory leaks exist.

rcoreilly:
Yeah, would it make sense to delete this PR for the time being and rebase on the variadic version which I will test now.  The variadic one does not have any leaks?

rcoreilly:
once you have it working with out leaks I mean.

nishantsharma:
Sure. Variadic shouldn't have leaks. I may take up fixing of memory leaks later as per internal priority.

nishantsharma:
Actually, it may still make sense to review from the perspective of behavior. Esp. translating error return into exceptions. Also, deleting PRs causes context to be lost. Will suggest making it a draft PR.

rcoreilly:
ok.. I'll try to take a look in a few days.  Meanwhile please run your code through gofmt -- that check is failing on test.

rcoreilly:
ps. thanks for submitting these fixes -- much appreciated!

nishantsharma:
URL changes are temporary. Made them to test. Will revert before switching PR draft status. JFYI.

PR #284, opened 2022-06-24
Add optional dynamic linking when building shared library

pyalex:
Currently shared library is built with static linking to python making the resulting library not portable.

This PR proposes flag "dynamic-link" (false by default) that would impact flags used on linking stage (passed to `ld`).
When this flag is enabled instead of linking to python (via `-l python3.x)` we will use flags similar to what python itself uses during building extensions (see `distutils.sysconfig.get_config_var("LDSHARED")`). This allows packing the resulting `.so` / `.dylib` file into python package and distribute it via PyPi.

Signed-off-by: pyalex <moskalenko.alexey@gmail.com>


pyalex:
This has been already tested during publishing [feast](https://github.com/feast-dev/feast) python package.

rcoreilly:
Looks good -- thanks for your contribution!

PR #285, opened 2022-06-28
Release GIL in function generator

pyalex:
When python callback is passed as an argument to a Go code, it's expected to work from any subsequently created go routine. Unfortunately, that's currently impossible due to GIL acquired and hold by the main go thread (in which Go was initially called).

This PR adds GIL release logic to the function generator. Before stepping into Go code, GIL will be released and then reacquired back.

Signed-off-by: Oleksii Moskalenko <moskalenko.alexey@gmail.com>

### DIFF
--- a/bind/gen_func.go
+++ b/bind/gen_func.go
@@ -260,6 +260,14 @@ func (g *pyGen) genFuncBody(sym *symbol, fsym *Func) {
 			}
 		}
 	}
+
+	// release GIL
+	g.gofile.Printf("_saved_thread := C.PyEval_SaveThread()\n")
+	if !rvIsErr && nres != 2 {
+		// reacquire GIL after return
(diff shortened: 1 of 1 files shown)


rcoreilly:
This GIL stuff is out of my limited python expertise but I assume you've tested, and this is easily disabled if there are problems, so I'll merge. Does it impact performance?

pyalex:
Hi @rcoreilly, as my previous PR this was as well tested within `feast` first. We have set of end-to-end tests that relies on this functionality and we also published at least two releases with this code and haven't heard complains so far.
I'm not sure what you referring to regarding performance. GIL doesn't impact Go code in any way except blocking parallel Python code execution.

rcoreilly:
Great!  I was just wondering about the overhead of those save / restore calls which weren't otherwise being called..

PR #286, opened 2022-06-28
feat: Add support for build tags

achals:
Signed-off-by: Achal Shah <achals@gmail.com>

We need support for build tags when trying to link to a project that has a dependency on certain build flags (in our case, arrow).


achals:
Hi @rcoreilly , would really appreciate a review on this when you had the chance!

rcoreilly:
yep looks good -- thanks for your contribution!

Issue #236, opened 2020-11-15
Add []error to standard types in go package


rcoreilly:
`bind/stdtypes.go` installs the basic builtin types, and Slice_ versions of these.  It does not currently install an []error slice, but it should -- otherwise functions returning this end up binding to the wrong version of it somehow.  I tried adding it but there was a crash during generation. Need to figure this out:

```
--- Processing package: github.com/emer/vision/vxform ---
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1246fa7]

goroutine 1 [running]:
go/types.(*Package).Name(...)
	/usr/local/Cellar/go/1.15.2/libexec/src/go/types/package.go:34
github.com/go-python/gopy/bind.(*symbol).pyPkgId(0xc0000d51d0, 0xc0000bb2c0, 0x2d, 0x0)
	/Users/oreilly/go/src/github.com/go-python/gopy/bind/symbols.go:361 +0x47
github.com/go-python/gopy/bind.(*pyGen).genSliceInit(0xc002623a40, 0xc0000dc960, 0x0, 0x0)
	/Users/oreilly/go/src/github.com/go-python/gopy/bind/gen_slice.go:191 +0x55eb
github.com/go-python/gopy/bind.(*pyGen).genSlice(0xc002623a40, 0xc0000dc960, 0xc0000a0000, 0x0)
	/Users/oreilly/go/src/github.com/go-python/gopy/bind/gen_slice.go:53 +0x3fb
github.com/go-python/gopy/bind.(*pyGen).genType(0xc002623a40, 0xc0000dc960, 0x0)
	/Users/oreilly/go/src/github.com/go-python/gopy/bind/gen_type.go:51 +0x1c5
github.com/go-python/gopy/bind.(*pyGen).genGoPkg(0xc002623a40)
	/Users/oreilly/go/src/github.com/go-python/gopy/bind/gen.go:854 +0x208
github.com/go-python/gopy/bind.(*pyGen).genPkg(0xc002623a40, 0xc0000d8000)
	/Users/oreilly/go/src/github.com/go-python/gopy/bind/gen.go:603 +0x105
github.com/go-python/gopy/bind.(*pyGen).gen(0xc002623a40, 0xc002623a40, 0x1341ef0)
	/Users/oreilly/go/src/github.com/go-python/gopy/bind/gen.go:524 +0xc5
[...]

PR #288, opened 2022-07-04
support error slices

kjahed:
Fix issue #236 by treating error as a basic type.


rcoreilly:
Thanks for the contribution!

PR #296, opened 2022-09-16
gen symbol name use index to avoid illegal symobl name

darren:
when conflicting symbols exceeds, the generated symbol name will have control character.


rcoreilly:
sorry for the long delay -- just getting a break here to do these -- thanks for the fix!

PR #297, opened 2022-11-13
Fix - slice element is pyobject

NoamK-CR:
When a slice element is a `PyObject*` pybindgen needs to know the ownership details.


rcoreilly:
looks good - -thanks!

Issue #301, opened 2022-12-09
Building multiple go packages produces wrong python3 imports


psha-:
This is a leftover from #245

PR #302, opened 2022-12-09
Fixes a leftover bug to properly handle multiple go packages when gen…

psha-:
Fixes a leftover bug to properly handle multiple go packages when generating python relative imports.

Refs go-python/gopy #245


psha-:
Fixes #301

rcoreilly:
builds fine for me on my mac -- not sure what is going on with the missing package on CI.. nothing that would have changed with this PR though.

PR #310, opened 2023-02-01
Fix: mixed absolute and relative imports

### LINKED ISSUE #309: Mixed relative and absolute imports
When running `gopy` according to the docs and having multiple packages in the module where one imports the other, relative as well as absolute imports are used. Thereby, it becomes impossible to import the depending package.

Exemplary project structure:

```
mymod
|-- ecs
|   `-- ecs.go
`-- model
    `-- model.go
```

Where `model` imports `ecs`.

Running this command:

```
gopy pkg -output=out github.com/mlange-42/mymod
```

The generated code in `model.py` looks like this:

```python
# ...

from ..mymod import _mymod
from ..mymod import go

os.chdir(cwd)

# to use this code in your end-user python file, import it as follows:
# from mymod import model
# and then refer to everything using model. prefix
# packages imported by this package listed below:

from mymod import ecs

# ...
```

Depending on where I start the REPL and how I try to import, I get either:

```
    from ..mymod import _mymod
ImportError: attempted relative import beyond top-level package
```

(Run from `out`, with `from mymod import model`)

or:

```
    from mymod import ecs
ModuleNotFoundError: No module named 'mymod'
```

(Run from `mymod`, with `from out.mymod import model`)

### Workaround

The only way I could get it to work (without editing generated Python files) is to run `gopy` with `package-prefix` set to nothing:

```
gopy pkg -output=out -package-prefix= github.com/mlange-42/mymod
```

Then, running the Python REPL from `out`, it works:

```python
from mymod import model
```

This is not obvious from the docs, and probably also not intendend.
Possibly related to #245, #301, #302

mlange-42:
## Problem

Command `pkg` creates a mix of absolute and relative imports that make using the Python package impossible. See #309.

## Solution

Handle imports for command `pkg` the same as `build` and `exe`.
For the default `PkgPrefix` (`.`), all imports now look like this:

```python
from . import xxx
```

## Notes

Not sure why it was implemented differently for `pkg` compared to `build` and `exe`, so my approach may not be valid. However, it could also be a leftover bug from #245, #301 and #302.

Fixes #309

### DIFF
--- a/bind/gen.go
+++ b/bind/gen.go
@@ -551,7 +551,7 @@ func (g *pyGen) genPkgWrapOut() {
 	// note: must generate import string at end as imports can be added during processing
 	impstr := ""
 	for _, im := range g.pkg.pyimports {
-		if g.mode == ModeGen || g.mode == ModeBuild {
+		if g.mode == ModeGen || g.mode == ModeBuild || g.mode == ModePkg {
 			if g.cfg.PkgPrefix != "" {
 				impstr += fmt.Sprintf("from %s import %s\n", g.cfg.PkgPrefix, im)
 			} else {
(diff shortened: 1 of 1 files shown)


rcoreilly:
yeah not clear why it is different -- did you run the tests?  I will look at later just to make sure..

PR #312, opened 2023-02-02
Fix: Broken with Go 1.20 - quoted path

### LINKED ISSUE #311: Broken with Go 1.20 - empty sections in Python output
With the new Go version 1.20 released yesterday, `gopy` is broken.

It produces only empty `.py` files, except for the preamble. I.e. all the `Types`, `Enums`, `Interfaces`, `Structs` etc. sections are emptly.
When I switch back to Go 1.19.5, it works as expected again.

I also observed that some warnings like these diappeared with the non-working 1.20:

```
ignoring python incompatible method: ... gopy: return type is signature
```

(i.e. the respective functions are probably not even processed)

EDIT/Further observations:

Under Go 1.20, the package types are already empty here:
https://github.com/go-python/gopy/blob/7a98aa58fe4927ea67c828f10af62b26c896478d/gen.go#L138

Also, when it comes to collecting symbols, the package is marked as incomplete (`complete=false`) in 1.20, while it is complete in 1.19.

mlange-42:
Remove quotes from path when compiling package.

I.e., go was called as `go build -v path` (with "path" instead of the actual path).

Broke package extraction in Go 1.20.
I have no idea why it worked at all in Go 1.19.

Fixes #311

### DIFF
--- a/gen.go
+++ b/gen.go
@@ -97,7 +97,7 @@ func loadPackage(path string, buildFirst bool, buildTags string) (*packages.Pack
 			buildTagStr := fmt.Sprintf("\"%s\"", strings.Join(strings.Split(buildTags, ","), " "))
 			args = append(args, "-tags", buildTagStr)
 		}
-		args = append(args, "-v", "path")
+		args = append(args, "-v", path)
 		fmt.Printf("go %v\n", strings.Join(args, " "))
 		cmd := exec.Command("go", args...)
 		cmd.Stdin = os.Stdin


rcoreilly:
thanks for the fix -- do we need to up the go.mod version of packages?

mlange-42:
> thanks for the fix -- do we need to up the go.mod version of packages?

Works for me as is. Quite new to Go, so not sure what the implications would be.

rcoreilly:
Just remembered that the path has to be quoted so spaces in path names (common on windows) still work.  So you'll want to revert that one.

rcoreilly:
I can fix the go.mod when I pull it and run tests etc -- `go get -u <package>` is the command to get the updated latest version.

mlange-42:
> Just remembered that the path has to be quoted so spaces in path names (common on windows) still work. So you'll want to revert that one.

Ah, ok... but not quoted as it was, which just added `path` as argument. We want `"<path>"`. Will fix it.

Tests aren't in the CI? Looks so to me... 🤔

mlange-42:
Quotes are not supported here:

```
go build -v "github.com/mlange-42/gopy_ecs"
panic: path "\"github.com/mlange-42/gopy_ecs\"" not in error "invalid import path \"\\\"github.com/mlange-42/gopy_ecs\\\"\"" [recovered]
```

But thinking a bit further, they are also not required. An array of arguments is passed, which already considers the spaces/quotes that would come from the command line. I.e. the path is one element in the array, not multiple due to splitting by spaces.

rcoreilly:
good point!  Yeah our CI is broken -- @sbinet was using that appveryor thing -- eventually I'll transition to just use the github one which should work..

mlange-42:
Looks like some Python 2 stuff is failing to install. Also, the `Install Linux packages` step looks quite complicated. Is this really all necessary? And is Python 2 support really still required?

tamirfri:
@rcoreilly don't you ashamed? the project is broken. and the awesome guy need to wait more than a month to get this damn hotfix in. meanwhile I just forked the repo of @mlange-42. shame on you and on the sleepy maintainers

sbinet:
@tamirfri I am not sure your comment is really trying to be constructive.
we (well, at least, I, when I was sending some of my free cycles this way) are all working on this "pro-bono".
we are all stretched thin.

perhaps a simple, polite, _ping_ would have been better suited to reinject momentum in this PR.

rcoreilly:
@tamirfri it does take a bit of time to test the updates locally and make the release version, and I had some major deadlines.  I know it is frustrating and I've been waiting for other maintainers to do merges too, so I can see both sides of this..

PR #321, opened 2023-03-31
cmd: remove dependency on sed

### LINKED ISSUE #266: 'sed' is not recognized as an internal or external command...
sed is not a windows utility and shouldn't be used. I had to build in git bash after handling a handful of other issues to get a package to build. Why not just use python to do it?

sbinet:
Fixes #266.


rcoreilly (approved):
I hope I didn't do this in the wrong order.. we'll see..

PR #324, opened 2023-04-14
Enforce the extension module in setup.py

vmarkovtsev:
Without this change, [cibuildwheel](https://github.com/pypa/cibuildwheel) cannot build a correct wheel from the generated project. The problem is that the native lib is added by `MANIFEST.in` and `python3 -m pip wheel` ignores it and thinks that the package is pure Python. Therefore it builds a pure wheel and the packaging pipeline crashes.

### DIFF
--- a/pkgsetup.go
+++ b/pkgsetup.go
@@ -19,6 +19,12 @@ const (
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
+
+class BinaryDistribution(setuptools.Distribution):
+    def has_ext_modules(_):
+        return True
+
+
 setuptools.setup(
     name="%[1]s%[2]s",
     version="%[3]s",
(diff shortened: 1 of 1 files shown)


rcoreilly (approved):
cibuildwheel looks very useful -- thanks for the fix!

vmarkovtsev:
btw I also needed the following: ```
patchelf --remove-needed libpython3.11.so.1.0 path/to/gopy/built/_library.cpython-*.so
```
this way auditwheel doesn't cry on missing libpython. Note that this dependency is indeed redundant in case the native extension is imported in already running Python interpreter.

rcoreilly:
this is something you do in cibuildwheel config?  not something to add here, right?  You could potentially add that somewhere in README.md if so.

vmarkovtsev:
Well, it depends. Ideally, `gopy` wouldn't produce shared libraries linked to `libpython*`. Every time we import the wrapper, Python is already running by definition so the missing symbols resolve without having to link against `libpython`. I can dive deeper into how `gopy` works and offer a patch.

PR #338, opened 2023-10-03
fix __next__ function

richecr:
Previously, it was not possible to iterate using the `for ... in ...` command in Python as shown below:

```python
# Library generated by gopy
from hltvsdk import operations

# `matches` is of the slice type
matches = operations.GetMatches()

for m in matches:
    print(m.Team1)
```

The code above resulted in the following error:

```
Traceback (most recent call last):
  File "/home/richecr/open_source/hltvgo/test.py", line 6, in <module>
    print(m.Team1)
          ^^^^
AttributeError: 'int' object has no attribute 'Team1'
```

My fix addresses this issue. The solution essentially involved replicating the code generation as performed in the `__getitem__` method.

### DIFF
--- a/bind/gen_slice.go
+++ b/bind/gen_slice.go
@@ -246,7 +246,11 @@ otherwise parameter is a python list that we copy from
 		g.pywrap.Indent()
 		g.pywrap.Printf("if self.index < len(self):\n")
 		g.pywrap.Indent()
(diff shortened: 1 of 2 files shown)


richecr:
@rcoreilly hi, can you review this PR?

rcoreilly:
thanks for the contribution & sorry for the delay in merging.

richecr:
No problem, I thank you for the library made available.

PR #341, opened 2023-10-26
Generate const documentation.

### LINKED ISSUE #340: No consts documentation is generated
Py doc should be generated for const definitions such as:

```
// This is a const
const myConst1 = 1

// This is a typed const
const myConst myType = 1
```
Go handles such comments and it should not be a problem for gopy to generate py doc.
However it appears that go itself can not handle docs for const defined in a common const construct:

```
const (
    // Doc that can't be handled by go atm
    myConst1 = 1
)
```

psha-:
Generate const documentation. Additionally added typed const doc, besides untyped const doc.
Fixes #340.

### DIFF
--- a/bind/gen_varconst.go
+++ b/bind/gen_varconst.go
@@ -125,6 +125,16 @@ func (g *pyGen) genConstValue(c *Const) {
 		val = "False"
 	}
 	g.pywrap.Printf("%s = %s\n", c.GoName(), val)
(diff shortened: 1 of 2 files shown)


psha-:
@rcoreilly, can we merge this?

rcoreilly:
@psha- sorry for the delay and thanks for the contribution!  I'll cut a release now too.

Issue #328, opened 2023-05-02
Best way to pass data chunks (bytes, bytearray)?


morgenroth:
What is the best way to pass (python -> go) and retrieve (python <- go) data chunks?

I have tried to create a simple interface as shown below, but this does not compile into a working interface.

```golang
func DataWrite(handle int, buf []byte) int {
...
}

func DataRead(handle int, buf []byte) int {
...
}
```

gopy seems to handle `[]byte` as kind of object and then `.handle` is missing.
```
AttributeError: 'bytes' object has no attribute 'handle'
```

Then I tried to go with the Slice approach but aside from the drawback mentioned in #323, I have not found any way to actually put data into the arrays.

Can somebody provide or point me to a working example (go and python code) which shows how arrays of bytes can be efficiently passed with gopy? Thanks!

rcoreilly:
This is really a dupe of #323.  don't explicitly create the handle -- gopy manages all that.  slices example should be sufficient, also see the python wrapper to see exactly what that looks like.

Issue #323, opened 2023-04-03
Allow for fetching entire `[]byte` slice as `bytes`


deuill:
Currently, `[]byte` slices in Go are presented in Python as a custom type, `Slice_byte`, which allows for iterating over and fetching individual byte values from the underlying slice. However, this is rather inefficient for large slices, as each iteration requires FFI calls between Python, C, and Go (i.e. against the `Slice_byte_elem` function).

It might, therefore, be more efficient to allow for returning the entire `[]byte` slice as a Python `bytes` object, as a single FFI call, perhaps as a `C.CString`. Even if each (or the first) call creates a copy, the latency and CPU time tradeoff should be a sufficient improvement overall.

rcoreilly:
This could be achieved by using a string type presumably?  The semantics of the []byte would be lost if it was not directly writable on the python side.

deuill:
My understanding is that a `string` type in Go would automatically map to an `str` type in Python, which is defined as being composed only of UTF-8 bytes. Indeed, trying something similar out seems to return an error on the autogenerated `Get` method:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

I suppose there's no way of forcing a Go `string` to be handled as a `bytes` type instead? My assumption here is that there's not (or that any conversion here would be of equivalent effort as the original ask, in GoPy itself), and it's also partially the reason why I specified `bytes` rather than `bytearray` -- I assume that mapping the mutating aspects of the Python `bytearray` back to the Go `[]byte` may not be feasible, hence it might simply be worth adding a function that returns an immutable copy of the entire byte slice instead.

rcoreilly:
I'm not sufficiently up on the relevant standards in Python for how this all works, so I can't really judge, but it sounds like we might want to have it work in different ways depending on the use case.. We do have the ability to flag things with some kind of comment directive I believe, so that might be an option.  I can't quite remember where this is used but I believe it determines how an `interface{}` is treated or something to that effect.

rcoreilly:
[...]

PR #342, opened 2023-11-19
Efficient copies of bytes

NoamK-CR:
This aims to solve both #323 and #328 by supplying efficient methods to copy Python `bytes` into Go `[]byte` slices and vice versa.

One question I have is whether the `Slice_byte` Python constructor should check if the first argument is of type `bytes`, and if so use the optimized method?

### DIFF
--- a/SUPPORT_MATRIX.md
+++ b/SUPPORT_MATRIX.md
@@ -11,6 +11,7 @@ _examples/consts | yes
 _examples/cstrings | yes
 _examples/empty | yes
 _examples/funcs | yes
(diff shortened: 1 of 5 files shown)


rcoreilly:
this looks interesting -- but will take more time than I have avail right now.  also can you fix the build after I update with other PRs?

NoamK-CR:
Sure, I'll fix the build now and rebase.

rcoreilly:
Thanks and sorry for the delay in merging!

Issue #346, opened 2024-01-29
Can't return a slice of slices ([][]bool)


moritz157:
I wanted to pass a slice of slices of boolean values ([][]bool). I would assume that this could simply be done like this:
```go
func GetEmptyMatrix(xSize int, ySize int) [][]bool {
	result := [][]bool{}

	for i := 0; i < xSize; i++ {
		result = append(result, []bool{})
		for j := 0; j < ySize; j++ {
			result[i] = append(result[i], false)
		}
	}

	return result
}
```

In Python I then simply called the function:
```python
from out import matrix

matrix.GetEmptyMatrix(4,4)
```

However, this failed with the following error: `panic: interface conversion: interface {} is []bool, not *[]bool`

Is passing a slice of slices generally not supported or do I have to do it in a different way (like passing a slice of pointers to slices)? If this is supported, I would suggest to maybe add an example for this because as far as I can see, there is none yet. If I'm just doing it wrong, a hint in the right direction would be greatly appreciated :)

rcoreilly:
I'm assuming this is now fixed by #350 unless someone tells me otherwise.  And if someone could write a test to confirm, that would be most appreciated (@EvanOman ?)

EvanOman:
@rcoreilly No, my previous PR did not fix slices, only maps

However I just added #351 which fixes this issue and adds a test based on the example from @moritz157

Issue #329, opened 2023-05-17
Panic when function returns a map of non-pointer structs


chalkytoast:
A function returning a map of struct values causes a panic:

`panic: interface conversion: interface {} is batman.Foo, not *batman.Foo`

```
type Foo struct {
	S string
}

// crash
func Bar() map[string]Foo {
	return map[string]Foo{"foo": Foo{"bar"}}
}

// no crash
func Baz() map[string]*Foo {
	return map[string]*Foo{"foo": &Foo{"bar"}}
}
```

richecr:
I believe I have this same problem. I'm studying trying to make a sql query builder for Python using gopy, but I'm having a similar problem:

```
panic: interface conversion: interface {} is map[string]interface {}, not *map[string]interface {}
```

Function causing the error:
```
func (qb *QueryBuilder) Exec() ([]map[string]interface{}, error) {
	qb.Compiler.SetOptionsBuilder(qb.Statements, qb.Simple)
	result, err := qb.Compiler.Exec()
	qb.Reset()
	return result, err
}
```

qb.Compiler.Exec() function:
```
func (qc *QueryCompiler) Exec() ([]map[string]interface{}, error) {
	query := qc.Simple.Raw.Sql
	if query == "" {
		query = qc.ToSQL()
	}
	qc.Reset()

	var result []map[string]interface{}

	if qc.Simple.IsDQL {
		rows, err := qc.Client.Query(query)
		if err != nil {
			return nil, err
		}
		defer rows.Close()

		cols, err := rows.Columns()
		if err != nil {
			fmt.Println(err)
		}
		colTypes, err := rows.ColumnTypes()
		if err != nil {
			fmt.Println(err)
			return nil, err
		}

		vals := make([]interface{}, len(cols))
		for i, ct := range colTypes {
			switch ct.DatabaseTypeName() {
			case "VARCHAR", "TEXT":
				vals[i] = new(string)
			case "INT":
				vals[i] = new(int)
			default:
				vals[i] = new(interface{})
			}
		}

		for rows.Next() {
			scanArgs := make([]interface{}, len(cols))
			for i := range vals {
				scanArgs[i] = &vals[i]
			}

			err = rows.Scan(scanArgs...)
			if err != nil {
				fmt.Println(err)
				continue
			}

			rowMap := make(map[string]interface{})
			for i, colName := range cols {
				valPtr := vals[i]

				var val interface{}
				switch v := valPtr.(type) {
				case *string:
					if v != nil {
[...]

PR #350, opened 2024-04-22
Fix `handleFromPtr` missing Reference in Maps

EvanOman:
This patch fixes a `panic` I was encountering when using a map whose value was a non-value-type like another Map, Struct, Slice, etc.

Related:

- #346 
- #329 
- [Commit ](https://github.com/richecr/gopy/commit/92d159f53618f6df6857069e49109b16058e1f66) by @richecr that fixed a similar issue for slices

This fix makes `gopy` useable for me but I have not added new tests (yet)


richecr (approved):

rcoreilly:
Thanks for the fix and review.  Looks sensible to me.

PR #351, opened 2024-04-23
Adds missing pointer reference for slices along with test

### LINKED ISSUE #346: Can't return a slice of slices ([][]bool)

EvanOman:
Fixes #346 

NOTE: The variadic test was failing for me before adding this test:

<details>
  <summary>Click me</summary>

```
run cmd: [build -vm=python3 -output=/tmp/gopy-3250023487 -package-prefix  ./_examples/variadic]
go build -v ./_examples/variadic

--- Processing package: github.com/go-python/gopy/_examples/variadic ---

--- building package ---
gopy.test -test.paniconexit0 -test.timeout=10m0s
goimports -w variadic.go
go build -mod=mod -buildmode=c-shared -o variadic_go.so .
/home/evan/miniconda3/bin/python3 build.py
CGO_CFLAGS="-I/home/evan/miniconda3/include/python3.10" -fPIC -Ofast
CGO_LDFLAGS="-L/home/evan/miniconda3/lib" "-lpython3.10" -lcrypt -lpthread -ldl -lutil -lm -lm
go build -mod=mod -buildmode=c-shared -o _variadic.cpython-310-x86_64-linux-gnu.so .
running python3 test.py
FAIL
exit status 1
FAIL    github.com/go-python/gopy       123.954s
```

</details>

Is that test flaky for other people too?

However I did confirm that the new test I added failed before the fix and passed afterwards.

### DIFF
--- a/_examples/slices/slices.go
+++ b/_examples/slices/slices.go
@@ -60,3 +60,16 @@ func CmplxSqrt(arr SliceComplex) SliceComplex {
 	}
 	return res
 }
+
+func GetEmptyMatrix(xSize int, ySize int) [][]bool {
+	result := [][]bool{}
+
+	for i := 0; i < xSize; i++ {
+		result = append(result, []bool{})
+		for j := 0; j < ySize; j++ {
+			result[i] = append(result[i], false)
+		}
+	}
+
+	return result
+}
--- a/_examples/slices/test.py
+++ b/_examples/slices/test.py
@@ -44,4 +44,11 @@ for root, orig in zip(sqrts, cmplx):
     assert math.isclose(root_squared.real, orig.real)
     assert math.isclose(root_squared.imag, orig.imag)
 
+
+matrix = slices.GetEmptyMatrix(4,4)
+for i in range(4):
+    for j in range(4):
+        assert not matrix[i][j]
+print("[][]bool working as expected")
+
 print("OK")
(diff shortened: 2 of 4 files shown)


richecr (approved):
Thank you very much for the correction, I had this error.
I'm eagerly waiting for the release.

rcoreilly:
Giving this a try now.

rcoreilly:
looks like it works -- not sure how I missed the test before. Need to fix the CI testing here too someday.  just had to run format on your slices.go file.

PR #353, opened 2024-04-30
Replacement c.long(size) for c.longlong(size)

Inotart:
修复 python 3.12 版本无法编译的Bug
原生成后代码:
```golang
func Slice_byte_to_bytes(handle CGoHandle) *C.PyObject {
	s := deptrFromHandle_Slice_byte(handle)
	ptr := unsafe.Pointer(&s[0])
	size := len(s)
	return C.PyBytes_FromStringAndSize((*C.char)(ptr), C.long(size))
}
```
替换此例子中的C.long为C.longlong
实测修复后gopy build 正常

### DIFF
--- a/bind/gen_slice.go
+++ b/bind/gen_slice.go
@@ -406,7 +406,7 @@ otherwise parameter is a python list that we copy from
 			g.gofile.Printf("s := deptrFromHandle_Slice_byte(handle)\n")
 			g.gofile.Printf("ptr := unsafe.Pointer(&s[0])\n")
 			g.gofile.Printf("size := len(s)\n")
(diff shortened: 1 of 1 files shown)


rcoreilly:
thanks for the fix -- makes sense!

rcoreilly:
actually, this gives all kinds of errors on python 3.11:

```
./hi.go:304:53: cannot use _Ctype_longlong(size) (value of type _Ctype_longlong) as _Ctype_long value in argument to (_Cfunc_PyBytes_FromStringAndSize)
```

so I guess this is a 3.12 vs 11 difference?

PR #355, opened 2024-05-11
unconditional use strings.TrimSuffix

testwill:


rcoreilly (approved):
Yeah that's better - thanks!

PR #374, opened 2025-06-13
Add freebsd to main_unix.go

pswilde:
Add freebsd to main_unix.go, as without it building will fail as libExt and extraGccArgs are undefined


rcoreilly:
Thanks for the contribution!

PR #378, opened 2025-08-01
Update Versions in GH Actions

coffeemakingtoaster:
Lowest of the low hanging fruit, but this should make the GitHub action output a bit more useful. 
Didn't touch appveyor because I know too little about it.

Also bumped go version 1.22 to most recent (major) release

PR #379, opened 2025-09-04
Fix compilation with C23

deuill:
GCC 15 has been released (fairly) recently, and with it come C23 features such as built-in `bool` types, which obviate the need for defining these in each code-base.

In our case, this also causes compilation errors with GCC 15 (unless the `-std=gnu17` flag is passed), e.g.:

```
./test.go:16:17: error: two or more data types in declaration specifiers
   16 | typedef uint8_t bool;
      |                 ^~~~
```

This commit wraps these `typedef` declarations in `__STDC_VERSION__` checks, skipping if the version is C23 or over.

### DIFF
--- a/bind/gen.go
+++ b/bind/gen.go
@@ -52,7 +52,9 @@ package main
 %[3]s
 // #define Py_LIMITED_API // need full API for PyRun*
 #include <Python.h>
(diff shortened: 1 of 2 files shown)


deuill:
Tests will fail until #378 is merged (and this rebased).

deuill:
This was merged separately, closing...

PR #393, opened 2026-05-13
Fix memory leaks & crashes when loading multiple Go extensions in one Python process

b-long:
Relates-to: https://github.com/go-python/gopy/issues/392 
Fixes: https://github.com/go-python/gopy/issues/385
Fixes: https://github.com/go-python/gopy/issues/370

---

I took [Scusemua](https://github.com/Scusemua)'s two commits in https://github.com/go-python/gopy/pull/361 as a base and built a broader set of fixes on top, covering memory leaks, GC coordination, and shared library conflicts.

This PR wouldn't exist without the work Scusemua put in first. His two commits from July 2024 fixed a subtle but nasty bug: Go code was calling a Python function without holding the GIL.

That's a precise fix that prevents several hard-to-reproduce crashes, and everything else here builds on top of it.

---

I added four fixes in the same area:

1. C string memory leaks. Generated code allocated C strings when reading string fields from Go structs, slices, and maps, but never freed them. Switched those paths to a helper that frees after Python copies the value.

2. Unsafe GC call. The generated wrapper was calling Go's GC from a thread that crosses the C/Go boundary, which could panic on multi-core machines. Rerouted it through a background goroutine via a channel.

3. GC out of sync. Without coordination, Go heap objects piled up between Python GC cycles. Fixed by registering a gc.callbacks handler that triggers a Go collection after each Python one.

4. Symbol conflicts between co-loaded extensions. Loading two gopy-built shared libraries with RTLD_GLOBAL caused one library's Go internals to clobber the other's, corrupting heap state. Fixed by loading without that flag so each library's runtime stays isolated.

5. I added a regression test for two extensions in a single Python process, stealing from the discussion in issue 370. 

I also expanded the CI matrix to cover more Go versions and macOS configs.

Issue #370, opened 2025-04-07
"broken" since go 1.21


asottile:
I say "broken" because according to the golang issue this was never intended to work and was merely working by accident: https://github.com/golang/go/issues/65050#issue-2074509727

here's a small reproduction demoing this breakage.  I'm running on macos x86_64 which seems to be the most reliable to reproduce for some reason I don't completely understand:

**setup:**

```console
$ # not shown: running the install steps in the README
$ gopy build -output=hi -vm=python3 github.com/go-python/gopy/_examples/hi
go build -v github.com/go-python/gopy/_examples/hi

--- Processing package: github.com/go-python/gopy/_examples/hi ---

--- building package ---
gopy build -output=hi -vm=python3 github.com/go-python/gopy/_examples/hi
goimports -w hi.go
go build -mod=mod -buildmode=c-shared -o hi_go.so .
/Users/asottile/workspace/z/venv/bin/python3 build.py
CGO_CFLAGS="-I/Library/Frameworks/Python.framework/Versions/3.9/include/python3.9" -fPIC -Ofast
CGO_LDFLAGS="-L/Library/Frameworks/Python.framework/Versions/3.9/lib" "-lpython3.9" -ldl -framework CoreFoundation
go build -mod=mod -buildmode=c-shared -o _hi.cpython-39-darwin.so .
$ gopy build -output=simple -vm=python3 github.com/go-python/gopy/_examples/simple
go build -v github.com/go-python/gopy/_examples/simple

--- Processing package: github.com/go-python/gopy/_examples/simple ---

--- building package ---
gopy build -output=simple -vm=python3 github.com/go-python/gopy/_examples/simple
goimports -w simple.go
go build -mod=mod -buildmode=c-shared -o simple_go.so .
/Users/asottile/workspace/z/venv/bin/python3 build.py
CGO_CFLAGS="-I/Library/Frameworks/Python.framework/Versions/3.9/include/python3.9" -fPIC -Ofast
CGO_LDFLAGS="-L/Library/Frameworks/Python.framework/Versions/3.9/lib" "-lpython3.9" -ldl -framework CoreFoundation
go build -mod=mod -buildmode=c-shared -o _simple.cpython-39-darwin.so .
```

**reproduction**

here's a small python script using the compiled output:

(all that really needs to be involved is >1 extension in the same process)

```python
from hi.hi import Hello
from simple.simple import Add

for _ in range(5000):
    Add(2, 2)
    Hello('hi')
```

**boom**

[...]

PR #398, opened 2026-07-23
bind: make _gopy_clear_go_tls opt-in, off by default

### LINKED ISSUE #395: #370's _gopy_clear_go_tls (movq $0, %fs:-8) hard-crashes CPython 3.12+ in a shipping extension
Thanks for gopy. This is a follow-up to #370 and its fix, PR #393 ("Fix memory leaks & crashes when loading multiple Go extensions in one Python process", merged 2026-06-02), which introduced `_gopy_clear_go_tls()`. That workaround is now a hard, deterministic crash for a real, released project. slidge-whatsapp (a WhatsApp/XMPP gateway, https://codeberg.org/slidge/slidge-whatsapp) segfaults CPython on the first call into its gopy extension. Full downstream investigation: https://codeberg.org/slidge/slidge-whatsapp/issues/153

PR #393 solved a real multi-extension case, but the TLS clear is applied unconditionally, and in the single-extension case (which is the common one) it does the opposite: it corrupts CPython's own thread state. Concrete details below; I am glad to test any patch.

## Reproduction (real project, deterministic, no network/credentials)

```
python3.12 -m venv venv
venv/bin/pip install slidge-whatsapp==0.4.0b1      # ships a gopy-built wheel
venv/bin/python -c "import slidge_whatsapp.generated.whatsapp as w; w.NewGateway()"
```

```
Segmentation fault (core dumped)      # exit 139, every time, on the very first call
```

Reproduced on Gentoo, glibc, distro CPython 3.12 / 3.13 / 3.14 (identical), go 1.26.x, with gopy pinned at `v0.4.11-0.20241206185020-5f285b890023`. It does NOT crash on slidge-whatsapp 0.3.11b1, which predates `_gopy_clear_go_tls` (zero occurrences of it, of `RequestGC`, `_gcReq`, `gc.callbacks`); every build after it crashes.

## Backtrace

```
Thread 1 "python" received signal SIGSEGV
#0  _PyType_GetDict               (libpython3.12.so.1.0)
#1  _PyObject_GenericGetAttrWithDict
#3  PyObject_GetAttr
#4  _PyEval_EvalFrameDefault
...
si_addr = 0x10
rip     = _PyType_GetDict+48:  mov 0x10(%rdx),%rcx
rdx     = 0x0                   # NULL
```

`rdx` is the current thread-state / interpreter pointer. It is NULL right after the CGo call returns and Python bytecode resumes, so `0x10(%rdx)` faults at address 0x10 (the "segfault at 10" seen in dmesg).

## Root cause

The generated Python wrapper calls `_whatsapp._gopy_clear_go_tls()` before every CGo entry. On linux/amd64 its body is:

```c
static void _gopy_clear_go_tls(void) {
#elif defined(__x86_64__) && defined(__linux__)
    __asm__ volatile("movq $0, %%fs:-8" ::: "memory");
#endif
}
```

CPython 3.12 moved the current-thread-state fast path into a thread-local variable (`_Py_tss_tstate`, a `_Py_thread_local PyThreadState *`). It lives in the static TLS block, i.e. at a negative offset from the `%fs` thread pointer that the linker/loader assigns. The comment on `_gopy_clear_go_tls` assumes `%fs:-8` is Go's private "current g" slot; that is true for Go's internal TLS layout, but when Go is built `c-shared` and loaded through glibc, the g pointer is reached via a linker-assigned TLS variable (`runtime.tls_g`), not a fixed `-8`. glibc owns the TCB and the static TLS block; Go's g slot and CPython's `_Py_tss_tstate` both live there at loader-chosen offsets. Writing a hardcoded `%fs:-8` therefore clobbers whatever thread-local sits at that offset: here, CPython's current thread state, so `_PyThreadState_GET()` returns NULL and the interpreter crashes.

## Why it slips through CI ("works on my machine")

The colliding offset depends on the exact static-TLS layout, so the same `.so` crashes on one CPython build and not another. slidge-whatsapp's CI builds the wheel against a standalone CPython (the cgo preamble hardcodes `-I/root/.local/share/uv/python/...`), which is also CI's runtime, and there `%fs:-8` does not collide. Users install that same wheel against their distro CPython, where it does. That is exactly why [#153](https://codeberg.org/slidge/slidge-whatsapp/issues/153) kept getting "cannot reproduce" while some users on a distro Python hit it.

## Proof it is this instruction

Overwriting the single `movq $0, %fs:-8` (bytes `64 48 c7 04 25 f8 ff ff ff 00 00 00 00`, 13 bytes) with NOPs in the exact crashing `.so` removes the crash entirely; the extension then runs under real traffic. I fixed it for my particular installation using a post-build patch, but I believe it is a workaround for something that should be handled in gopy.

## Possible directions (suggestions)

1. Make the TLS clear opt-in, or skip it unless more than one gopy runtime is actually present, so the single-extension case (probably the majority) is safe by default. 
2. Drop the hardcoded `%fs:-8` / `%gs:0x30`. In `c-shared` mode the g pointer is at a linker-assigned offset (`runtime.tls_g`), so `-8` is not guaranteed to be Go's slot.
3. If the intent is to force `needm()` for a foreign thread, prefer a mechanism that does not poke raw TLS from C (e.g. pin with `runtime.LockOSThread` on the entry goroutine, or let `needm()` establish context per entry).

Happy to build test binaries or run patches against the slidge-whatsapp repro above.

satarsa:
## Summary

The `_gopy_clear_go_tls()` call added for issue #370 runs unconditionally before every CGo entry and performs a hardcoded TLS store (`movq $0, %fs:-8` on linux/amd64, `movq $0, %gs:0x30` on darwin/amd64). On glibc + CPython 3.12+ that offset overlaps the TLS slot CPython uses for the current thread state (`_Py_tss_tstate`), so the store nulls it and the interpreter segfaults on the first call: in the common **single-extension** case, which structurally cannot have the cross-runtime collision the clear was meant to guard against.

This PR makes the clear **opt-in** via a new `-clear-go-tls` build flag (default `false`): the default build is safe, and the previous behavior stays one flag away.

Fixes #395.

## The two problems with the unconditional clear

1. **It runs for a single extension too.** That is the common case and the one that cannot have a cross-runtime TLS collision, so there the clear is pure downside.
2. **The address is wrong for `c-shared` under glibc.** In a normal Go binary the runtime lays out its TLS so the g pointer sits at `%fs:-8`. In a `c-shared` library loaded by glibc the loader owns the static TLS block and the g pointer is reached through a linker-assigned TLS variable (`runtime.tls_g`), not a fixed `-8`. So `movq $0, %fs:-8` does not clear Go's g at all: it zeroes whatever thread-local happens to sit there, which on CPython 3.12+ is `_Py_tss_tstate`.

PR #393 already shipped the mechanism that actually isolates the runtimes: each extension is loaded without `RTLD_GLOBAL`, and gopy links each `.so` with a version script exporting only `PyInit_*`. Both keep every Go runtime symbol (including the g-pointer TLS) local to its own `.so`, so there is nothing shared to collide over. The raw-TLS clear is therefore redundant, and with the wrong offset actively harmful.

## What this PR changes

- Adds a `bind.ClearGoTLS` option and a `-clear-go-tls` build flag (default `false`), mirroring the existing `-no-make` flag across `cmd_build.go`, `cmd_gen.go`, `cmd_pkg.go`, `cmd_exe.go`.
- Gates the single `_gopy_clear_go_tls()` emission site in `bind/gen_func.go` on that flag.
- Leaves the C helper and its pybindgen registration in place, so `-clear-go-tls` restores today's behavior exactly.
- Leaves the `RTLD_GLOBAL` handling untouched and unconditional.

6 files, +33/-8. `gofmt`, `go vet ./...`, and `go test ./bind/` are clean.

## Verification

**Downstream (the crashing project).** Gentoo, glibc, CPython 3.14.6, go 1.26.4,
gcc 16. Rebuilding slidge-whatsapp's extension:

- Default (clear off): the generated wrapper contains zero `_gopy_clear_go_tls()` sites and the first call succeeds (exit 0).
- `gopy build -clear-go-tls`: the call reappears (one per CGo entry) and the crash returns (SIGSEGV, exit 139).

So the flag is a faithful toggle, and off-by-default removes the crash with no post-build patching.

**Multi-extension case, clear off.** Two independent gopy extensions (two Go runtimes) loaded in one interpreter, 200000 interleaved cross-calls on the same OS thread with periodic `gc.collect()` (the condition #370 describes): 0 wrong results, no crash, no corruption. And each `.so` exports exactly one global dynamic symbol (`PyInit_*`) and zero `runtime.*` symbols, so the g pointer is per-runtime-private regardless of `RTLD_GLOBAL`.

I do not have the original #370 reproducer, so I cannot claim to have re-run that exact case; the opt-in keeps the old behavior available for anyone who needs it.

## Follow-up (optional)

Happy to add regression tests, here or in a follow-up:

1. A codegen assertion (layout-independent): default flags produce a wrapper without `_gopy_clear_go_tls()`, `-clear-go-tls` produces one with it.
2. A two-extension `_examples` coexistence example, cross-called under `gc.collect()`, asserting a clean run with the clear off.

### DIFF
--- a/bind/gen.go
+++ b/bind/gen.go
@@ -329,12 +329,13 @@ cwd = os.getcwd()
 currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
 os.chdir(currentdir)
 # When multiple gopy extensions coexist in one Python process each carries its own
-# independent Go runtime. The Go extension is loaded without RTLD_GLOBAL below, and
-# _gopy_clear_go_tls() is called before each CGo entry to force needm() to run, which
-# establishes the correct per-extension M/P/mcache context (issue #370).
-# Also load the extension without RTLD_GLOBAL so that Go runtime symbols stay
-# local to each .so — belt-and-suspenders on platforms where RTLD_GLOBAL is the
-# Python default (e.g. some Linux builds).
+# independent Go runtime. Loading each extension without RTLD_GLOBAL below keeps its
+# Go runtime symbols (including the per-runtime goroutine-pointer TLS slot) local to
+# its own .so, which is what prevents the runtimes from colliding (issue #370).
+# A _gopy_clear_go_tls() call before each CGo entry is available as an extra safety
+# net but is opt-in (gopy build -clear-go-tls), off by default: on glibc + CPython
+# 3.12+ its hardcoded TLS store clobbers the interpreter thread state and crashes on
+# the first call in the common single-extension case (issue #395).
 if hasattr(sys, 'getdlopenflags'):
 	try:
 		import ctypes as _gopy_ctypes
@@ -517,6 +518,17 @@ var NoWarn = false
 
 // NoMake turns off generation of Makefiles
 var NoMake = false
+
+// ClearGoTLS controls whether generated wrappers emit a _gopy_clear_go_tls()
+// call before every CGo entry point (issue #370). It is opt-in and off by
+// default. The clear performs a hardcoded TLS store (movq $0, %gs:0x30 on
+// darwin/amd64, movq $0, %fs:-8 on linux/amd64); with a single gopy extension
+// in the process it is unnecessary, and on glibc + CPython 3.12+ that offset
+// overlaps the TLS slot CPython uses for the current thread state, so the store
+// nulls it and the interpreter segfaults on the first call (issue #395).
+// Loading each extension without RTLD_GLOBAL, done unconditionally, already
+// keeps each runtime's goroutine-pointer TLS local to its own .so.
+var ClearGoTLS = false
 
 // GenPyBind generates a .go file, build.py file to enable pybindgen to create python bindings,
 // and wrapper .py file(s) that are loaded as the interface to the package with shadow
--- a/bind/gen_func.go
+++ b/bind/gen_func.go
@@ -339,8 +339,13 @@ if __err != nil {
 	// Clear the Go TLS goroutine slot before the CGo entry point so that
 	// Go's needm() runs and establishes the correct per-extension context.
 	// Without this, two extensions sharing the same process can corrupt
-	// each other's heap via TLS collision (issue #370).
-	g.pywrap.Printf("_%s._gopy_clear_go_tls()\n", pkgname)
+	// each other's heap via TLS collision (issue #370). Opt-in and off by
+	// default: the hardcoded TLS store crashes CPython 3.12+ in the common
+	// single-extension case (issue #395), and RTLD_GLOBAL-local loading
+	// already isolates each runtime's goroutine-pointer TLS.
+	if ClearGoTLS {
+		g.pywrap.Printf("_%s._gopy_clear_go_tls()\n", pkgname)
+	}
 
 	// pywrap output
 	mnm := fsym.ID()
--- a/cmd_build.go
+++ b/cmd_build.go
@@ -45,6 +45,7 @@ ex:
 	cmd.Flag.Bool("symbols", true, "include symbols in output")
 	cmd.Flag.Bool("no-warn", false, "suppress warning messages, which may be expected")
 	cmd.Flag.Bool("no-make", false, "do not generate a Makefile, e.g., when called from Makefile")
+	cmd.Flag.Bool("clear-go-tls", false, "emit a _gopy_clear_go_tls() call before every CGo entry (issue #370); off by default, needed only when several gopy extensions share one process and known to crash CPython 3.12+ (issue #395)")
 	cmd.Flag.Bool("dynamic-link", false, "whether to link output shared library dynamically to Python")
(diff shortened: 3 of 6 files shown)


b-long (approved):
@satarsa Thanks so much for this PR! 🙇

I'm happy to merge this—it's definitely a great improvement for gopy. While reviewing, the most straightforward test I could think of was adding Python 3.12.x to the build matrix. (We clearly need to expand our test coverage, but that’s a separate topic for later.)

Would you mind taking a look at [PR #399](https://github.com/go-python/gopy/pull/399)? If you check the order of my commits, you can see where tests pass and fail. By cherry-picking and then reverting your patch, it seems to confirm just how valuable your work is for our Python 3.12 support.

I also ran into a bug while working on that PR and think I’ve patched it, but I’d love to get your thoughts on the fix if you have a moment.

Thanks again!

PR #399, opened 2026-07-24
feat: add '3.12' testing & improve complex conversion

### LINKED ISSUE #400: Complex-returning functions crash under CPython 3.12
Functions returning complex64 or complex128 crash on their first call under CPython 3.12 (x86_64 and arm64). CPython 3.11 is unaffected.

The crash is inside PyComplex_FromDoubles (bind/gen.go), a segfault consistent with a null PyThreadState.

The issue can be observed in Github Actions output, within https://github.com/go-python/gopy/actions/runs/30058888912

b-long:
This PR is serving two goals:

1. It regression tests https://github.com/go-python/gopy/pull/398 , by enabling Python 3.12 builds
2. It identifies and fixes (commit d41db33699c93239969e74e9afe1effe006c802b ) issue #400 

While we shouldn't combine the two, I'm doing so to be pragmatic and move quickly (for now).  Depending on feedback, I may split this PR into 2 separate parts.

Fixes: https://github.com/go-python/gopy/issues/400

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,13 +18,14 @@ env:
 jobs:
 
   build:
-    name: Build
+    name: Build (${{ matrix.platform }}, Go ${{ matrix.go-version }}, Python ${{ matrix.python-version }})
     strategy:
       fail-fast: false
       matrix:
         # TODO: Consider official support matrix (OS and Go versions) and adjust this matrix accordingly
         go-version: [1.25.x, 1.24.x, 1.23.x, 1.22.x]
         platform: [ubuntu-latest, windows-latest, macos-15]
+        python-version: ['3.11', '3.12']
     runs-on: ${{ matrix.platform }}
     steps:
     - name: Checkout code
@@ -33,7 +34,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.11'
+        python-version: ${{ matrix.python-version }}
 
     - name: Install Go
       uses: actions/setup-go@v5
--- a/bind/gen.go
+++ b/bind/gen.go
@@ -329,12 +329,13 @@ cwd = os.getcwd()
 currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
 os.chdir(currentdir)
 # When multiple gopy extensions coexist in one Python process each carries its own
-# independent Go runtime. The Go extension is loaded without RTLD_GLOBAL below, and
-# _gopy_clear_go_tls() is called before each CGo entry to force needm() to run, which
-# establishes the correct per-extension M/P/mcache context (issue #370).
-# Also load the extension without RTLD_GLOBAL so that Go runtime symbols stay
-# local to each .so — belt-and-suspenders on platforms where RTLD_GLOBAL is the
-# Python default (e.g. some Linux builds).
+# independent Go runtime. Loading each extension without RTLD_GLOBAL below keeps its
+# Go runtime symbols (including the per-runtime goroutine-pointer TLS slot) local to
+# its own .so, which is what prevents the runtimes from colliding (issue #370).
+# A _gopy_clear_go_tls() call before each CGo entry is available as an extra safety
+# net but is opt-in (gopy build -clear-go-tls), off by default: on glibc + CPython
+# 3.12+ its hardcoded TLS store clobbers the interpreter thread state and crashes on
+# the first call in the common single-extension case (issue #395).
 if hasattr(sys, 'getdlopenflags'):
 	try:
 		import ctypes as _gopy_ctypes
@@ -517,6 +518,17 @@ var NoWarn = false
 
 // NoMake turns off generation of Makefiles
 var NoMake = false
+
+// ClearGoTLS controls whether generated wrappers emit a _gopy_clear_go_tls()
+// call before every CGo entry point (issue #370). It is opt-in and off by
+// default. The clear performs a hardcoded TLS store (movq $0, %gs:0x30 on
+// darwin/amd64, movq $0, %fs:-8 on linux/amd64); with a single gopy extension
+// in the process it is unnecessary, and on glibc + CPython 3.12+ that offset
+// overlaps the TLS slot CPython uses for the current thread state, so the store
+// nulls it and the interpreter segfaults on the first call (issue #395).
+// Loading each extension without RTLD_GLOBAL, done unconditionally, already
+// keeps each runtime's goroutine-pointer TLS local to its own .so.
+var ClearGoTLS = false
 
 // GenPyBind generates a .go file, build.py file to enable pybindgen to create python bindings,
 // and wrapper .py file(s) that are loaded as the interface to the package with shadow
--- a/bind/gen.go
+++ b/bind/gen.go
@@ -335,12 +335,13 @@ cwd = os.getcwd()
 currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
 os.chdir(currentdir)
 # When multiple gopy extensions coexist in one Python process each carries its own
-# independent Go runtime. The Go extension is loaded without RTLD_GLOBAL below, and
-# _gopy_clear_go_tls() is called before each CGo entry to force needm() to run, which
-# establishes the correct per-extension M/P/mcache context (issue #370).
-# Also load the extension without RTLD_GLOBAL so that Go runtime symbols stay
-# local to each .so — belt-and-suspenders on platforms where RTLD_GLOBAL is the
-# Python default (e.g. some Linux builds).
+# independent Go runtime. Loading each extension without RTLD_GLOBAL below keeps its
+# Go runtime symbols (including the per-runtime goroutine-pointer TLS slot) local to
+# its own .so, which is what prevents the runtimes from colliding (issue #370).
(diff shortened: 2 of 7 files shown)


satarsa:
Thanks for merging #398! Wiring up the 3.12 matrix is exactly the coverage this needed, I'm happy to have a look. I've reproduced #400 locally end-to-end before writing this, so below is what I've found, a confirmation that your fix is correct, and one structural observation you may want to weigh for later.

As I now understand, #400 is a real bug, fully independent of #398. Your `PyGILState_Ensure` / `Release` around `PyComplex_FromDoubles` is correct and I verified it fixes the crash. The root cause is that gopy releases the GIL (`PyEval_SaveThread`) around the whole Go wrapper body, and the complex conversion is the one path that *allocates* a Python object inside that GIL-released window. That's why only complex crashes and, e.g., int or `bytes` do not.

## Environment

Gentoo, glibc, system CPython **3.14.6**, go 1.26.4, gcc 16, x86_64. Note 3.12 is only the *floor*: this class of crash is the glibc + CPython-3.12+ layout, not a specific minor, so 3.14 reproduces #400 identically. Your CI hitting it on 3.12 and me hitting it on 3.14 are the same bug.

## 1. #400 is real and orthogonal to #398

I built `gopy` from master-with-#398 (clear **off** by default) and ran a small probe package, each function in its own process (a fault shows up as a non-zero/signal exit):

| function | Go->Py conversion path | result (clear off, no complex fix) |
|---|---|---|
| `Add(i, j int) int` | C side (pybindgen) | OK, `5` |
| `GetString() string` | C side, `C.CString` | OK, `hello` |
| `RealOf(complex128) float64` | Go side, `PyComplex_AsCComplex` (read, no alloc) | OK, `3.0` |
| `GetBytes() []byte` | C-side handle (`go.Slice_byte`) | OK |
| `bytes(GetBytes())` | Go side, `PyBytes_FromStringAndSize` | OK, `b'\x01\x02\x03\x04'` |
| `MayFail(int) (int, error)` | value + raise | OK / `RuntimeError` |
| `Comp64Add(complex64) complex64` | Go side, `PyComplex_FromDoubles` (alloc) | **SIGSEGV** |

So with the TLS clear already off, complex still crashes while everything else is fine; this is not #395 wearing a complex hat, it is a distinct bug. Confirmed crash signature matches your report:

```
Fatal Python error: Segmentation fault
Current thread (most recent call first):
  File ".../simple.py", line 107 in Comp64Add
```

## 2. Root cause: allocation inside gopy's GIL-released window

The generated wrapper releases the GIL for the entire body:

```go
//export probe_Comp64Add
func probe_Comp64Add(i *C.PyObject, j *C.PyObject) *C.PyObject {
	_saved_thread := C.PyEval_SaveThread()        // GIL released, current tstate nulled on this thread
	defer C.PyEval_RestoreThread(_saved_thread)   // re-acquired on return
	return complex64GoToPy(probe.Comp64Add(complex64PyToGo(i), complex64PyToGo(j)))
	//     ^^^^^^^^^^^^^^^^ PyComplex_FromDoubles allocates here, GIL-released -> null tstate
}
```

On CPython 3.12+, object allocation dereferences the current thread state (allocator / GC bookkeeping), so allocating with the GIL released and `tstate == NULL` faults. This explains the whole table above:

- int / float / bool / string are marshaled on the **C side** by pybindgen, where the GIL is still held -> fine.
- `[]byte` comes back as a **handle** (C side); only an explicit `bytes(x)` hits the Go-side `PyBytes_FromStringAndSize`: and that runs as its **own** exported call with the GIL held, so it's fine too.
- `PyComplex_AsCComplex` (reading a complex **arg**) runs in the released window but only reads two doubles, no allocation -> fine.
- `PyComplex_FromDoubles` (building the complex **result**) is the one Go-side **allocation** inside the released window -> crash.

## 3. As I see your fix is correct and verified

`PyGILState_Ensure()` re-acquires the GIL and binds a valid tstate for exactly the allocation, `Release()` drops it again. I rebuilt gopy with your two-function patch and:

```
Comp64Add(1+2j, 3+4j) -> (4+6j)     # exit 0
RealOf(3+4j)          -> 3.0        # exit 0
Add(2,3)              -> 5          # exit 0
```

It also matches gopy's own precedent: the Python-callback path in `bind/symbols.go` already wraps its C-API calls in `PyGILState_Ensure` / `Release` for the same reason. So this is consistent with the existing design, not a one-off.

## 4. Two a little stinky places, they're optional and may be for later

1. **The arg-read side is still GIL-released.** `complex64PyToGo` / `complex128PyToGo` (`PyComplex_AsCComplex`) touch a `PyObject` while the GIL is released. It doesn't fault (no allocation), but strictly, touching any `PyObject` without the GIL is UB. So the complex round-trip as a whole runs Python-C-API-against-released-GIL; the allocation is just the part that visibly crashes.

2. **The tuple builder.** `bind/symbols.go` has a Go-side `PyTuple_New` / `gopy_build_*` path. I could not get it onto a live wrapper path from ordinary signatures (plain multi-return like `(int,int)` is rejected: "second result value must be of type error"; `(T, error)` returns value+raise, no tuple), so I could not make it crash. But if any construct emits that builder inside the released window, it's the same bug.

The minimal, robust generalization would probably be to **not** `PyEval_SaveThread` for wrappers whose signature marshals `*C.PyObject` on the Go side (complex today), or to keep the GIL for the marshalling prologue/epilogue and only release around the pure-Go call. Your targeted `PyGILState_Ensure` is a perfectly good fix for the crash right now; the above is just the fuller shape of it if you ever want to close the window entirely.

b-long:
@satarsa Thank you so much for the detailed feedback 🙇 

I stacked a PR on top of this one, and I wonder if you'd be generous enough to look at it as well: https://github.com/go-python/gopy/pull/401

One thing to note: I tested locally (with the changes from PR 401) and reverting d41db33699c93239969e74e9afe1effe006c802b on top of changes in 401. I expected that would be enough to resolve the original segfault, since 401 moves the complex arg marshalling ahead of `PyEval_SaveThread` — but it isn't. `TestBindSimple` segfaults on the very first `Comp64Add` call, same fault signature as the original report.

Tracing it: 401 only fixes the *argument* side (your point 1 above — `complex64PyToGo`/`128PyToGo` now run before the GIL is released). The crash fixed in d41db33 is on the *return* side. `complex64GoToPy`/`128GoToPy` get called from inside `return complex64GoToPy(pkg.Comp64Add(...))`, and since `PyEval_RestoreThread` is only `defer`red, Go evaluates that whole expression — including the allocation — before the deferred restore runs. So the `PyGILState_Ensure`/`Release` added there is still the only thing protecting it; 401 doesn't touch or replace it.

So it looks like d41db33 and the two commits in 401 are covering two different, adjacent windows (return-side allocation vs. argument-side read, plus the tuple builder from your point 2) rather than one overlapping one. As far as I can tell, we need all of the above 🙂

satarsa:
I would merge this as is.

I checked the tree at `202b0fb` rather than the commit list, since apply, revert, then merge master looks alarming from the outside: `var ClearGoTLS = false` (bind/gen.go), the `if ClearGoTLS` gate (bind/gen_func.go) and the preamble comment are all present, so reverting the branch-local copy did not undo what landed in master. CI is green on all 24 jobs.

Two things worth considering:

1. **3.12 is the floor of this bug class, not the ceiling.** Everything in #400 reproduces identically on 3.14, which is where I hit it. As written the matrix will keep catching the bug you already fixed and will not catch the next one. Adding `'3.13'` and `'3.14'` to `python-version` is a one-line change, but note it takes the matrix to 4 x 3 x 4 = 48 jobs. If that is more CI than you want to spend, an alternative is keeping the full cross product on 3.11 and using `include:` to add the newer Pythons on ubuntu only, since the failure mode is glibc + CPython layout rather than platform specific.

2. **Squash when merging**, so that `Revert "bind: make _gopy_clear_go_tls opt-in, off by default"` does not land in master's history. It is accurate within the branch and misleading everywhere else, and it is exactly the kind of thing that costs someone half an hour during a bisect a year from now.

b-long:
> I would merge this as is.

Yep, I agree and will do. Thank you for the detailed collaboration. It's been a tremendous help! 🙇 

> 3.12 is the floor of this bug class, not the ceiling. 

Absolutely 👍 My thinking is we ought to rework the test matrix into something that is careful with the budget, and simultaneously more correct.  I don't have strong feelings about how, but certainly we should improve the current situation.

A sequence like the one below could keep us from spending unnecessarily, while also increasing coverage 🤷 

```yaml
name: Python Sequence Test Example

on: [push, pull_request]

jobs:
  test-3-11:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Set up Python 3.11
        uses: actions/setup-python@v5
        with:
          python-version: "3.11"
      - run: ...

  test-3-12:
    needs: test-3-11
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Set up Python 3.12
        uses: actions/setup-python@v5
        with:
          python-version: "3.12"
      - run: ...

  test-3-13:
    needs: test-3-12
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Set up Python 3.13
        uses: actions/setup-python@v5
        with:
          python-version: "3.13"
      - run: ...

  test-3-14:
    needs: test-3-13
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Set up Python 3.14
        uses: actions/setup-python@v5
        with:
          python-version: "3.14"
      - run: ...

```

PR #402, opened 2026-07-28
feat: Add gopy version command & release-please pipeline

### LINKED ISSUE #356: Cannot use ldflags to set build-time go variables
## Desired behaviour
When building a go module, it is useful to have a version string "burned" into the binary.
In go, this is accomplished by using the `go tool link` option `-X`, via the `go build` option `-ldflags`.
E.g.: `go build -ldflags="-X BUILD_VERSION=${VERSION}"`
That command tells the linker to use the value in the bash variable `VERSION` as the initial value for the go variable `BUILD_VERSION`.

## Current behaviour
Although the -ldflags line is used internally in the gopy code, and passed to the `go build` command, it is hard-coded to set just the `-s` and `-w` options, based on the gopy option `--symbols`.
https://github.com/go-python/gopy/blob/1d7f3a2d29bd9f3921ff640f8e4e8908170ff3bf/cmd_build.go#L193

## Possible solution
Add a new gopy build option `--ldflags` that would pass its value on to `go build`.
The `-w` and `-s` flags could be de-duplicated if the gopy `--symbols` is set, and those flags were also included in the `--ldflags` value.

## Development
I plan to fork the project and fix this for my own requirements shortly, and would be grateful if a contributor could revierw the change once tested, and see whether they consider it a useful enhancement, to be merged into the main project.

b-long:
Explicitly manages `ldflags`, stamps version into the CLI output, and automates GitHub releases.

Fixes https://github.com/go-python/gopy/issues/356


cmd_version.go
```diff
@@ -0,0 +1,32 @@
+// Copyright 2015 The go-python Authors.  All rights reserved.
```
b-long:
```suggestion
// Copyright 2026 The go-python Authors.  All rights reserved.
```

